PR #29, opened 2019-12-03
Facets

StuartHarris:
In order to present an alphabetical list to allow users to drill down into active substances and associated products and their documents, we need to be able to use the facets capability of the search engine. This means that we need to store a collection of (hierarchical) facets in the metadata associated with each SPC/PIL (PARs don't have active substances).

We can then construct from these facets the UI that will allow users to navigate in this way

[Airtable ticket](https://airtable.com/tbl7CISHKkr8Kdeh2/viwlDIU6cp8cbQYhj/recQRU7eH8rYPrKTE?blocks=hide)


StuartHarris:
Temporarily removed deploy job from pipeline until we have permissions to set up a service connection

PR #37, opened 2019-12-04
Change clickable area of search results

### LINKED ISSUE #34: Only Title on Search Result should be clickable
Currently, when we search for documents, the entire box of the result is clickable and leads to the document. We should change this so that only the title is clickable.

m-doughty:
### Link to Github issue

resolves #34 

### Acceptance Criteria

- [x] Only the title of search results should take you to the document.

### Testing information

1. Navigate to the website.
2. Perform a valid search which returns results (`ibuprofen` for example).
3. Clicking on the title (or file size) should load the document, clicking anywhere else should not.

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/web/components/search-results/index.tsx
+++ b/medicines/web/components/search-results/index.tsx
@@ -26,12 +26,14 @@ const StyledDrugList = styled.section`
     margin: 0;
   }
 
-  ul {
+  dl,
+  dt,
+  dd {
     list-style: none;
     padding: 0;
   }
 
-  li {
+  article {
     display: flex;
     background-color: ${mhraBlue10};
     padding: ${baseSpace};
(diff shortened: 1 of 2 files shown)


medicines/web/components/search-results/index.tsx
```diff
@@ -165,26 +165,26 @@ const SearchResults = (props: { drugs: IDocument[]; lastSearch: string }) => (
               <p className="icon">{drug.docType.toUpperCase()}</p>
             </div>
             <div className="right">
-              <a href={drug.url}>
-                <p className="drug-name">
+              <p className="drug-name">
```
pataruco: Please, if this a title of a result item I suggest to change `<p>` for a heading tag, e.g `<h3>`

pataruco:
For Skateboard lite is OK, but I think a more semantically accurate tags for this would be [`<dl>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl)

medicines/web/components/search-results/index.tsx
pataruco: Resolved

m-doughty: Agreed, we should raise an issue to review semantic use of HTML but this is out of scope of this bugfix.

m-doughty:
@pataruco I've raised issue #39 so that we can ensure we're using correct semantic HTML going forward.

PR #54, opened 2019-12-06
Update issue templates

m-doughty:
These are the default GitHub issue templates. Should be a good starting point.


pataruco:
Very nice @m-doughty

PR #72, opened 2019-12-09
Add reporting to importer, and skip unreleased & duplicate files.

### LINKED ISSUE #48: Unreleased documents
Document metadata has a field showing the "release state". Either we should import released documents only, or we should filter out unreleased documents on all searches.

### LINKED ISSUE #49: Reports from import process
We should modify the importer so that it reports which documents it has seen before (i.e. duplicates) and for which documents it has no metadata. This will allow us to more easily compare source and target in order to ensure that the import is correct.

### LINKED ISSUE #77: Reference to the Agency in the footer
**Describe the bug**
The current wording says 'Built by the Medicines & Healthcare products Regulatory Agency'

The wording should say: 'Built by the Medicines and Healthcare products Regulatory Agency' - we only use the ampersand (the &) when in logo form, not when written out

**To Reproduce**

1. View the footer
iPad Pro landscape

m-doughty:
### Tickets

Resolves #48 
Resolves #49 
Resolves #77 

### Acceptance Criteria

- [x] Duplicate PARs are not uploaded;
- [x] Duplicate SPCs & PILs are not uploaded;
- [x] PARs with no metadata are not uploaded;
- [x] SPCs & PILs with no metadata are not uploaded;
- [x] PARs which are not released are not uploaded;
- [x] SPCs & PILs which are not released are not uploaded;
- [x] The importer returns a useful report at the end on STDOUT for PARs;
- [x] The importer returns a useful report at the end on STDOUT for SPCs & PILs.

**Note: I believe this is working for SPCs & PILs too but that needs testing, and the sample data has metadata in a different format to real data so I'll need @StuartHarris to test this for me (or I can download the SPC & PIL data).**

### Testing information

Run the importer.

### Pre-merge Checklist

- [ ] Branch test approved


medicines/import/src/report.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    pub(crate) fn add_skipped_unreleased(&mut self, file_name: &str, release_state: &str) {
+        self.skipped_unreleaseds.push(SkippedUnreleased {
+            file_name: file_name.to_string(),
+            release_state: release_state.to_string()
+        });
+    }
+
+    pub(crate) fn already_uploaded_file_with_hash(&self, hash: &str) -> bool {
+        self.uploaded.iter().any(|uploaded| uploaded.hash == hash)
```
StuartHarris: This is cool! And fine as we're not that performance sensitive, but was thinking that if we were, we could use a `HashMap` and check and add in one go:
```rust
use std::collections::HashMap;
let mut map = HashMap::new();
map.entry(hash).or_insert(true);
```

StuartHarris: Would probably want to match on the `Entry` so we can add to the report if `Entry::Occupied`. Anyway, not really important

m-doughty: Yeah, if we were dealing with a dataset of like 5 million I would definitely have gone for the hashmap here.

StuartHarris: cool

medicines/import/src/main.rs
```diff
@@ -11,22 +11,29 @@ use tokio_core::reactor::Core;
 fn main() -> Result<(), AzureError> {
     let yaml = load_yaml!("cli.yaml");
     let matches = App::from_yaml(yaml).get_matches();
+    let verbosity: i8;
+    match matches.occurrences_of("verbose") {
+        0 => verbosity = 0,
+        1 => verbosity = 1,
+        2 | _ => verbosity = 2,
+    };
```
StuartHarris: love this!

medicines/import/src/pdf.rs
```diff
@@ -6,7 +6,7 @@ use std::{
 pub fn get_pdfs(dir: &Path) -> io::Result<Vec<PathBuf>> {
     Ok(fs::read_dir(dir)?
         .filter_map(Result::ok)
-        .filter(|entry| entry.path().extension().map(|s| s == "pdf").is_some())
+        .filter(|entry| entry.path().extension().unwrap() == "pdf")
```
StuartHarris: thanks!

medicines/web/components/footer/index.tsx
```diff
@@ -78,9 +78,7 @@ const Footer: React.FC = () => (
           <li>
             <p>
               Built by the&nbsp;
-              <a href="">
-                Medicines &amp; Healthcare products Regulatory Agency
-              </a>
+              <a href="">Medicines and Healthcare products Regulatory Agency</a>
```
StuartHarris: I think you may get a conflict on this as it's been removed completely now!

PR #113, opened 2019-12-11
Search box

StuartHarris:
Updates the search box to run horizontally across the top and incorporate Active Substance drill down links. This means that these links stay visible even when doing a search. The new search box is "both of the ways that you can find results" and is always available at the top of the search results. As a result of this change there is no sidebar and the Yellow Card CTA is also horizontal across the top.

### DIFF
--- a/medicines/web/components/mip/index.tsx
+++ b/medicines/web/components/mip/index.tsx
@@ -158,6 +158,9 @@ const Mip: React.FC = () => {
           horizontal
         />
       </div>
+      <div className="yellow-card-wrapper">
+        <YellowCard />
+      </div>
       <div>
         {showingResultsForTerm.length === 0 ? (
           <>
@@ -181,9 +184,6 @@ const Mip: React.FC = () => {
           />
         )}
       </div>{' '}
-      <div className="yellow-card-wrapper">
-        <YellowCard />
-      </div>
     </StyledMip>
   );
 };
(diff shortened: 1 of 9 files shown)


medicines/web/components/mip/index.tsx
```diff
@@ -84,14 +57,6 @@ const Mip: React.FC = () => {
...
   };
 
-  const fetchFacetResults = async (searchTerm: string) => {
-    const searchResults = await facetSearch(searchTerm);
-    const filtered = searchResults[1].facets.filter(x =>
-      x.value.startsWith(searchTerm),
-    );
-    setFacetResults(filtered);
-  };
-
```
rjkerrison: 👍

medicines/web/components/search-results/index.tsx
```diff
@@ -239,7 +237,7 @@ const SearchResults = (props: {
                   </dt>
                   <dd className="right">
                     {drug.product != null ? (
-                      <a href={drug.url}>
+                      <a href={drug.url} download="x">
```
rjkerrison: Not sure why but this `download` attribute doesn't seem to do anything for me locally. I just navigate to the PDF as normal.

StuartHarris: good catch! I was testing out the HTML5 download attribute

medicines/web/components/search-results/index.tsx
```diff
@@ -249,6 +247,17 @@ const SearchResults = (props: {
                       </a>
                     )}
                     <p className="metadata">File size: {drug.fileSize} KB</p>
+                    <p className="metadata">Created: {drug.created}</p>
```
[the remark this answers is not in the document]
StuartHarris: thanks have removed now also

PR #120, opened 2019-12-11
remove intro on drill down

### LINKED ISSUE #86: Alphabetised Search - View Shifts to top of the Page
**Describe the bug**
When clicking on a Letter or number the View shifts to the top of the page and you need to scroll down to see the list of Active Substance returned

**Expected behaviour**
View to show the list of Active substances returned

### LINKED ISSUE #116: Accessibility - Adding a skip to results button on the homepage
**Is your feature request related to a problem? Please describe.**
The problem when filtering for active substances on the homepage is that the entire copy for describing MHRA, SPC-PILs and PARs displays before the results. For a user with a screenreader, there is a lot of copy they have to get through before reaching the relevant content. 

**Describe the solution you'd like**
It would be more accessible to have a button which allows users to skip to the results.
This would also benefit all users so they can get to the relevant content faster. 

**Describe alternatives you've considered**
Alternatively the copy describing MHRA, SPC-PILs and PARs could be removed when filtering for an active substance. 

**Additional context**

StuartHarris:
Resolves #116. Resolves #86

Improve showing and hiding of home page components for search and active substance drill down.

### DIFF
--- a/medicines/web/components/mip/index.tsx
+++ b/medicines/web/components/mip/index.tsx
@@ -67,10 +67,10 @@ const sanitizeTitle = (title: string | null): string => {
 
 const Mip: React.FC = () => {
   const [pageNumber, setPageNumber] = React.useState(1);
+  const [hasIntro, setHasIntro] = React.useState(true);
   const [resultCount, setResultCount] = React.useState(0);
   const pageSize = 10;
   const [results, setResults] = React.useState<IDocument[]>([]);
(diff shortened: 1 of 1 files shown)


DimplePatel:
Does this PR address https://github.com/MHRA/products/issues/116 and/or https://github.com/MHRA/products/issues/86 ?

StuartHarris:
@DimplePatel yes! Thanks. Have edited so they will close automatically

PR #189, opened 2019-12-13
Terraforming infrastructure

StuartHarris:
Initial attempt at Terraform scripts for provisioning Azure infrastructure for both Products and CPD.

See the README.md


infrastructure/environments/non-prod/variables.tf
```diff
@@ -0,0 +1,26 @@
...
+  description = "Azure region"
+  default     = "uksouth"
+}
+
+variable "RESOURCE_GROUP_PRODUCTS" {
+  description = "Products resource group name"
+  default     = "products"
+}
+
+variable "RESOURCE_GROUP_CPD" {
```
callerc1: We got rid of this so we can remove.

PR #217, opened 2019-12-17
Resize Yellow Card image in IE 11

craiga:
#### Before

Screenshot 2019-12-17 at 16 45 56

#### After

Screenshot 2019-12-17 at 16 45 46

### DIFF
--- a/medicines/web/components/yellow-card/index.tsx
+++ b/medicines/web/components/yellow-card/index.tsx
@@ -25,6 +25,9 @@ const StyledYellowCard = styled.section`
   header {
     padding: 1rem;
     background-color: ${mhraYellow};
+    svg {
+      height: 31px;
+    }
   }
 
   div.action-bar {


medicines/web/components/yellow-card/index.tsx
```diff
@@ -25,6 +25,9 @@ const StyledYellowCard = styled.section`
   header {
     padding: 1rem;
     background-color: ${mhraYellow};
+    svg {
+      height: 31px;
```
rjkerrison: would it change/break anything for this to be in `rem`?

craiga: No, but we know the image should be 31 pixels tall. I'd need to figure out how many `rem`s that is. In my not-very-strongly-held opinion, `px` makes more sense in this context, but happy to change if you like.

Issue #234, opened 2019-12-19
Incremental Imports of SPC/PIL/PAR Data


m-doughty:
**Is your feature request related to a problem? Please describe.**

We need to be able to import SPC/PIL/PAR data as it is released, and delete SPC/PIL/PAR data as it's redacted or unreleased, without incurring the downtime of a full index rebuild.

**Describe the solution you'd like**

The import tool should be modified to have an `incremental` mode which accepts two files, an input CSV and an output CSV. It will create a diff of the CSVs and create an internal `todo` list:

1. Files which are in the old CSV but not the new CSV will be marked for deletion;
2. Files which are in the new CSV but not the old CSV will be marked for upload;
3. Files which are in both CSVs with difference in metadata will be re-uploaded.

This may necessitate some changes to the existing behavior (looking at CSV before files) because it's easier to reason about the application if it works the same way in both modes.

**Describe alternatives you've considered**

We considered making a new container and new index each increment from scratch, but there were several issues with this approach:

1. The difficulty of moving data from Stellant;
2. The need to upgrade our Azure account;
3. The need to spin up cloud servers to run the account.

m-doughty:
After extensive discussions between @StuartHarris and myself, and this task growing ever more complicated with each new discovery about the old data format and metadata format, we don't believe that this is the correct way to handle the imports.

Instead, our intention is to do the following:

1. Upgrade the Azure search account so that we can have two indexes side by side;
2. Create two storage containers;
3. Redeploy all documents -- new and old -- every friday(?) when we get the new batch;
4. When the re-deploy is done, switch the site over to the new index/container.

This avoids a huge amount of complexity and insures that any given visitor is always seeing a consistent result set -- old or new.

PR #242, opened 2019-12-20
Incremental Import

m-doughty:
### Ticket

#234 

### Acceptance Criteria

- [x] We should be able to set an old CSV, a new CSV, and a JSON file of old hashes;
- [x] Files in the old CSV but not the new CSV should be deleted from blob storage;
- [x] Files in both CSVs where the metadata hasn't been changed should be skipped;
- [x] Files in both CSVs where the metadata has changed should be replaced;
- [x] Files in the new CSV but not the old CSV should be uploaded;
- [x] All of the above should be represented in the report.

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/import/Pipfile
+++ b/medicines/import/Pipfile
@@ -1,0 +1,11 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+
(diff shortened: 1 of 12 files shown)


m-doughty:
After a lot of discussion between myself and @StuartHarris, this isn't the direction we want to take this in.

PR #254, opened 2020-01-08
Filters

### LINKED ISSUE #370: SEARCH - Basic Filters (spc/pil/par)
### User want
As a user I want to be able to filter my search results on the search page, so that I can find what I'm looking for quicker and easier

1. I'm able to use a filter on the search results page and the category I've chosen gives me results I'm expecting
2. I am able to easily find and know how to use filters on the search results page

**Data acceptance criteria**
1. Filter events are tracked by category and label and we can cross reference that with what page a user was on when they filtered

1. Filtering works per the category a user has chosen to apply
2. Usage of a filter fires in GA under a custom event

**Size**
M

**Value**
8

**Effort**

### Exit Criteria met 
- [x] Backlog
- [x] Discovery
- [x] DUXD
- [x] Development 
- [x] Quality Assurance 
- [ ] Release and Validate

rjkerrison:
### Filters

Closes #370.

Damn fine coffee

### Acceptance Criteria

- [x] I'm able to use a filter on the search results page and the category I've chosen gives me results I'm expecting
- [x] I am able to easily find and know how to use filters on the search results page
- [x] Filter events are tracked by category and label and we can cross reference that with what page a user was on when they filtered
- [x] Filtering works per the category a user has chosen to apply
- [x] Usage of a filter fires in GA under a custom event

### Testing information

<insert notes for testers that might be helpful>

### Pre-flight Checklist

- [ ] Tests
- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


medicines/web/src/components/mip/index.tsx
```diff
@@ -119,17 +121,41 @@ const Mip: React.FC = () => {
     });
   };
 
+  const handleDocTypeCheckbox = async (d: DocType) => {
```
m-doughty: Can we use more explicit variable names? `docType` and `allowedDocTypes` for example?

medicines/web/src/components/mip/index.tsx
```diff
@@ -119,17 +121,41 @@ const Mip: React.FC = () => {
...
 
+  useEffect(() => {
+    rerouteSearchResults(1);
+  }, [docTypes]);
+
   useEffect(() => {
     if (searchTerm && page) {
       if (typeof searchTerm === 'string') {
         (async () => {
+          let d = null;
```
m-doughty:
```js
let docType = (typeof queryDocFilter === 'string') 
  ? formatDocTypeFilters(queryDocFilters) 
  : null;
```

medicines/web/src/components/search-filter/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+  checkDocType: (d: DocType) => void;
+}
+
+interface IDocTypeCheckboxProps extends ISearchFilterProps {
+  docType: DocType;
+  name: string;
+}
+
+const DocTypeCheckbox: React.FC<IDocTypeCheckboxProps> = props => {
+  const { docType: d, name, checkDocType, docTypes } = props;
```
m-doughty: `docType`

medicines/web/src/components/search-filter/index.tsx
```diff
@@ -0,0 +1,56 @@
...
+      />
+      <label htmlFor={id}>
+        {name} ({d.toUpperCase()})
+      </label>
+    </p>
+  );
+};
+
+const SearchFilter: React.FC<ISearchFilterProps> = props => {
+  const generateCheckbox = (d: DocType, n: string) => {
```
m-doughty: Readable variable names please.

medicines/web/src/services/azure-search.ts
```diff
@@ -98,6 +57,10 @@ const buildSearchUrl = (
   url.searchParams.append('search', query);
   url.searchParams.append('scoringProfile', searchScoringProfile as string);
   url.searchParams.append('searchMode', 'all');
+  const f = createFilter(filters);
```
m-doughty: `filter`

medicines/web/src/services/keyword.ts
```diff
@@ -0,0 +1,38 @@
...
+  'ig',
+);
+
+const escapeSpecialWords = (word: string): string =>
+  word.replace(/(\|\||&&|\bAND\b|\bOR\b|\bNOT\b)/gi, `\\$1`);
+
+const preferExactMatchButSupportFuzzyMatch = (word: string): string =>
+  `${word}~${searchWordFuzziness} ${word}^${searchExactnessBoost}`;
+
+const addNormalizedProductLicenses = (q: string): string => {
```
m-doughty: What is `q`?

medicines/web/src/components/mip/index.tsx
rjkerrison: Sure

medicines/web/src/components/search-filter/index.tsx
rjkerrison: Suffers from semantic satiation because everything here is named `docType`. It's a short-lived variable. Calling it `d` in this context makes it much quicker to scan where it's coming from, than looking for `docType`.

medicines/web/src/services/keyword.ts
rjkerrison: This is a direct move of the code from https://github.com/MHRA/products/commit/8024817194e579bd284146de01e4ee4b1eedb7f4

medicines/web/src/services/keyword.ts
```diff
@@ -0,0 +1,38 @@
+const searchExactnessBoost = process.env.AZURE_SEARCH_EXACTNESS_BOOST;
```
TimboTambo: This file name could be mistaken for the use of `keywords` relating to the field of the same name. I think it'd be better renamed to e.g. `search-query-normalizer` or `search-query-cleaner`.

medicines/web/src/components/search-filter/index.tsx
TimboTambo: I agree that `docType` is better to use here - it reduces the need to understand where it's coming from at all, whereas you would always have to check where `d` comes from to work out what it is. The other doc-like variables could be renamed to make them more explicit and avoid confusion - e.g. `docTypes` could be `selectedDocTypes` or `filteredDocTypes`.

medicines/web/src/services/keyword.ts
rjkerrison: Renamed

medicines/web/src/services/azure-search.ts
rjkerrison: extracted this out so I can do some descriptive naming in the function, never liked how big this method was getting

medicines/web/src/components/search-filter/index.tsx
rjkerrison: Went for much, much more explicitly naming which should make the code read more like English now

medicines/web/src/components/mip/index.tsx
rjkerrison: thanks, have refactored this a bit to use the ternary and use the result for a check on whether to call the `useState` setter afterwards

rjkerrison:
Some merge conflicts here and we could do with some tests.

medicines/web/src/components/search-results/index.tsx
```diff
@@ -113,18 +118,36 @@ const StyledDrugList = styled.div`
...
-  name: string;
-  product: string;
-  url: string;
-}
+  .column {
+    float: left;
+  }
+  .filter {
+    width: 40%;
+  }
```
TimboTambo: May be worth setting a max width on this - 40% on a wide screen is gonna look ments. Similarly but less importantly, a min width?

TimboTambo: Considered using something like flexbox to control widths more fluidly across different size screens and devices?

medicines/web/src/components/mip/index.tsx
```diff
@@ -119,17 +121,44 @@ const Mip: React.FC = () => {
...
+  };
+
   const rerouteSearchResults = (pageNo: number) => {
     router.push({
       pathname: router.route,
-      query: { search, page: pageNo },
+      query: {
+        search,
+        page: pageNo,
+        doc: enabledDocTypes.length > 0 ? enabledDocTypes.join(',') : null,
```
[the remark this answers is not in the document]
rjkerrison: I was expecting `null` to mean that the parameter will be omitted

medicines/web/src/components/search-results/index.tsx
rjkerrison: Using flex now. Switches direction to `column` in mobile view, and a width is no longer specified so the amount of real estate on a large screen isn't bizarre now.

craiga:
If you run a search, apply a filter, and then click a page number, the search results will not be filtered to the selected document type.

medicines/web/src/components/mip/index.tsx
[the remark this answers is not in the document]
rjkerrison: I've changed this now to be an empty string. It's in a file called `querystring-interpreter.ts`.

rjkerrison:
Decision with @roughprada and @sandracoen1405 regarding behaviour for adding a filter is that it should go back to page 1 when changing filters. This is inline with other search functionality elsewhere on the internet: notably, Google behaves this way. (cc @craiga)

PR #265, opened 2020-01-16
GraphQL schema with pagination

rjkerrison:
### Schema of GraphQL for API

[Airtable ticket](https://airtable.com/tbl7CISHKkr8Kdeh2/viwlDIU6cp8cbQYhj?blocks=hide)

Look at this 'graph

### Acceptance Criteria

- [ ] Substances available from graphql endpoint with dummy data
- [ ] Products available from graphql endpoint with dummy data

Note that
* This is foundational work
* Real searching functionality will be completed in later PRs
* Pagination will also follow

### Testing information

To get to the graphiql sandbox:

1. `cd /medicines/api`
2. `cargo run`
3. Navigate to http://127.0.0.1/graphiql

### Pre-flight Checklist

- [ ] Tests
- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/api/Cargo.toml
+++ b/medicines/api/Cargo.toml
@@ -1,17 +1,17 @@
 [package]
 name = "api"
 version = "0.1.0"
-authors = ["Stuart Harris <stuart.harris@red-badger.com>"]
+authors = ["Stuart Harris <stuart.harris@red-badger.com>", "Robin James Kerrison <robinjames.kerrison@red-badger.com>"]
 edition = "2018"
 
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-actix-web = "2.0.0"
 actix-rt = "1.0.0"
+actix-web = "2.0.0"
 env_logger = "0.7.1"
-serde = "1.0.103"
-serde_json = "1.0.44"
-serde_derive = "1.0.103"
-juniper = "0.14.2"
+juniper = { git = "https://github.com/graphql-rust/juniper" }
 listenfd = "0.3.3"
+serde = "1.0.103"
+serde_derive = "1.0.103"
+serde_json = "1.0.44"
--- a/medicines/api/Cargo.toml
+++ b/medicines/api/Cargo.toml
@@ -10,7 +10,8 @@ edition = "2018"
 actix-rt = "1.0.0"
 actix-web = "2.0.0"
 env_logger = "0.7.1"
-juniper = { git = "https://github.com/graphql-rust/juniper" }
+futures = "0.3.1"
+juniper = { git = "https://github.com/graphql-rust/juniper", features = ["async"] }
 listenfd = "0.3.3"
 serde = "1.0.103"
 serde_derive = "1.0.103"
--- a/medicines/api/Cargo.toml
+++ b/medicines/api/Cargo.toml
@@ -13,6 +13,7 @@ env_logger = "0.7.1"
 futures = "0.3.1"
 juniper = { git = "https://github.com/graphql-rust/juniper", features = ["async"] }
 listenfd = "0.3.3"
+reqwest = { version = "0.10.1", features = ["json"] }
 serde = "1.0.103"
 serde_derive = "1.0.103"
 serde_json = "1.0.44"
--- a/medicines/api/README.md
+++ b/medicines/api/README.md
@@ -10,11 +10,16 @@ If you're using homebrew, you should be able to install `rustup` by running:
 1. `brew install rustup-init`
 2. `rustup-init` and following the instructions
 
+You may be able to install `rustup` through your chosen package manager, if you use one.
+
+Otherwise, visit [rust-lang's `rustup` installation guide][rustup install], which has instructions for your operating system.
+For installing
+
 You should also have installed:
 
-- [Docker](https://docs.docker.com/install/)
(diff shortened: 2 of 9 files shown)


medicines/api/src/azure_search.rs
```diff
@@ -0,0 +1,84 @@
...
+    #[serde(rename="@odata.count")]
+    count: Option<i32>,
+}
+
+fn get_env(key: &str) -> String {
+    match std::env::var(key) {
+        Ok(b) => b,
+        Err(_) => "".to_owned()
+    }
+}
```
craiga: So this swallows errors? Should it be named `get_env_or_empty_string` to make its purpose more clear?

rjkerrison: Worth discussing. If we do let the error bubble up, it's swallowed. We may need to move this so that if the envrionment variables are missing, the server doesn't start, so we'll see healthchecks fail.

m-doughty: I think we should handle errors explicitly. An environment variable not being set is usually going to result in an un-usable API so we may as well fatal here.

medicines/api/src/azure_search.rs
```diff
@@ -0,0 +1,83 @@
...
+            ("highlight", "content"),
+            ("queryType", "full"),
+            ("@count", "true"),
+            ("@top", "10"),
+            ("@skip", "0"),
+            ("search", &search_term),
+            ("scoringProfile", "preferKeywords"),
+        ])
+        .build()
+        .unwrap();
```
m-doughty: Better to use `?` and let the error bubble up into the result?

medicines/api/src/azure_search.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    let r = client.execute(req).await;
```
m-doughty: You can replace from here to the bottom of the file with:

```rust
client
  .execute(req)
  .await?
  .json::<AzureSearchResults>()
  .await?
```

And I think it's quite a bit cleaner than matching just to rethrow the error you caught.

medicines/api/src/main.rs
```diff
@@ -19,14 +23,11 @@ async fn graphql(
     st: web::Data<Arc<Schema>>,
     data: web::Json<GraphQLRequest>,
 ) -> Result<HttpResponse, Error> {
-    let user = web::block(move || {
-        let res = data.execute(&st, &());
-        Ok::<_, serde_json::error::Error>(serde_json::to_string(&res)?)
-    })
-    .await?;
+    let res = data.execute_async(&st, &()).await;
```
m-doughty: `.await?`

medicines/api/src/substance.rs
```diff
@@ -0,0 +1,39 @@
...
+    fn name(&self) -> &str {
+        &self.name
+    }
+}
+
+pagination! {Substances, SubstanceEdge, Substance}
+
+pub async fn get_substances(first: i32) -> Substances {
+    let substances: [&str; 1000] = ["Ibuprofen"; 1000];
+    let e = substances
```
m-doughty: Just call this `edges` and pass it into `Substances` directly.

medicines/api/README.md
```diff
@@ -7,7 +7,7 @@ This is where the API rust server lives.
 You should have `rust` installed, ideally `rustup`.
 If you're using homebrew, you should be able to install `rustup` by running:
 
-1. `brew instal rustup-init`
+1. `brew install rustup-init`
 2. `rustup-init` and following the instructions
```
m-doughty: Would like to see instructions for getting `rustup` for non-Mac users, especially on an open source project.

medicines/api/src/product.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    let e = products
+        .map(|y| ProductEdge {
+            node: y,
+            cursor: "cursor".to_owned(),
+        })
+        .collect();
+
+    Some(Products {
+        edges: e,
```
rjkerrison: will do the rename of `e` to `edges` here too

medicines/api/src/azure_search.rs
rjkerrison: @craiga @m-doughty This now `panic`s when necessary environmental variables are not found. I also rearchitected so that this is called when building a global context, so starting up the server will fail quickly and loudly.

medicines/api/src/azure_search.rs
rjkerrison: thanks

medicines/api/src/azure_search.rs
rjkerrison: I think this variable splitting was a hangover from when I was `println`-ing last week to understand the async problem I was seeing. Good catch. I've tidied it now.

medicines/api/Cargo.toml
```diff
@@ -1,17 +1,19 @@
 [package]
 name = "api"
 version = "0.1.0"
-authors = ["Stuart Harris <stuart.harris@red-badger.com>"]
+authors = ["Stuart Harris <stuart.harris@red-badger.com>", "Robin James Kerrison <robinjames.kerrison@red-badger.com>"]
```
pataruco: 😎

PR #273, opened 2020-01-21
Testing strategy principles

rjkerrison:
### Testing Strategy

[Airtable ticket](https://airtable.com/tbl7CISHKkr8Kdeh2/viwlDIU6cp8cbQYhj/recIZevMtRJYzhOkO?blocks=hide)

Test Card F

### Acceptance Criteria

- [ ] Testing guidelines

### DIFF
--- a/docs/principles/testing.md
+++ b/docs/principles/testing.md
@@ -4,14 +4,14 @@
 
 End-to-end testing tests an entire product, service or application. Individual products should each have their own tests.
 
-### Cypress
+### Websites tested by Cypress
 
 Websites should be tested by Cypress.
 
 - `/medicines/web`
 - `/learning/web`
 
-### API
+### APIs tested by Jest
 
 GraphQL APIs should be tested by Jest.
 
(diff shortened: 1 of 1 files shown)


docs/principles/testing.md
```diff
@@ -0,0 +1,45 @@
+# Testing Strategy
+
+## End-to-End Testing
+
+End-to-end testing tests an entire product, service or application. Individual products should each have their own tests.
+
+### Cypress
```
craiga: Headings here are a mix of tools and the things those tools are testing. Maybe they should all be in the form "Web sites are tested with Cypress", "APIs are tested with Jest", etc.?

rjkerrison: Good call, I've made a change for that.

docs/principles/testing.md
```diff
@@ -0,0 +1,45 @@
+# Testing Strategy
+
+## End-to-End Testing
+
+End-to-end testing tests an entire product, service or application. Individual products should each have their own tests.
```
craiga: Should this mention that external dependencies (such as Azure search) aren't under test so should be mocked out? I'd anticipate that this will keep coming up as a question about our automated end-to-end tests.

Suggest something like the following:

> Where possible, external dependencies should be mocked out. This is because external dependencies can slow down these tests and make them brittle. Testing with external dependencies should be carried out manually.

docs/principles/testing.md
```diff
@@ -0,0 +1,45 @@
...
+
+GraphQL APIs should be tested by Jest.
+
+- `/medicines/api`
+
+## Unit Testing
+
+### Helpers tested by Jest
+
+JavaScript and Typescript helpers should be tested by Jest.
```
craiga: Should be "TypeScript" (this is *super* pedantic).

rjkerrison: It's amazing I did one capitalisation and not the other.

PR #279, opened 2020-01-28
Improved Privacy Settings

craiga:
### Improved Privacy Settings

[Airtable ticket](https://airtable.com/tbl7CISHKkr8Kdeh2/viwlDIU6cp8cbQYhj/recgMZISroTTc2Xd5)

giphy-3

### Acceptance Criteria

In order to comply with ICO requirements, we would need to have all non-essential cookies disable by default and ask/allow the users to enable them

- [ ] The cookies are disable by default when you land on the website
- [ ] When the user clicks "accept all cookies", the cookies became active
- [ ] The user can go to the cookie policy page at any time and disable cookies and identify all essential cookies detailed as the analytics ones.

### Testing information

There may be an issue with routing when disabling cookies (the page is reloaded as `/cookies/`, which results in a 404 on my dev environment). I need to do more testing.

### Pre-flight Checklist

- [ ] Tests
- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


medicines/web/src/components/cookie-form/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+          type="radio"
+          name="cookie"
+          id="cookie-off"
+          checked={!props.storageAllowed}
+          onChange={handleCookiesOff}
+          role="button"
+        />
+        Off
+      </label>
+    </StyledCookieForm>
```
craiga: Note that this isn't contained within a `<form>`. Including a form element triggered an error with `a11y`. Not sure if handling this by removing the `<form>` is a good idea.

medicines/web/src/hooks/index.tsx
```diff
@@ -25,3 +30,34 @@ export const useSessionStorage = (key: string, initialValue: any): any => {
...
+        value instanceof Function ? value(storedValue) : value;
+      setStoredValue(valueToStore);
+      window.localStorage.setItem(key, JSON.stringify(valueToStore));
+    } catch (error) {
+      // tslint:disable-next-line: no-console
+      console.error(error);
+    }
+  };
+  return [storedValue, setValue];
+};
```
craiga: I tried combining these into one function which worked something like this:

```javascript
const useStorage => (storageBackend: any) => {
  return (key: string, initialValue: any): any => {
    …
    return [storedValue, setValue];
  }
}

export const useSessionStorage = useStorage(window.sessionStorage);
export const useLocalStorage = useStorage(window.localStorage);
```

…but something doesn't like me using `window` in this context. 🤷‍♂

medicines/web/src/components/cookie-form/index.tsx
pataruco: What is the error?

And yes is very bad idea to remove `form` tag

medicines/web/src/components/page/index.tsx
```diff
@@ -1,7 +1,11 @@
 import Head from 'next/head';
 import React from 'react';
+import ReactGA from 'react-ga-gtm';
+import TagManager from 'react-gtm-module';
 import styled from 'styled-components';
 import { Normalize } from 'styled-normalize';
+
+import { useLocalStorage } from '../../hooks';
```
pataruco: Please remove `useLocalStorage` because is not been used

medicines/web/src/components/cookie-form/index.tsx
craiga: @pataruco It was an issue about there being no submit button on the form. I tried to find an ARIA attribute to indicate that the radio buttons *are* the submit buttons, but I couldn't find it.

medicines/web/src/components/cookie-form/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+    // Delete *everything*
+    window.localStorage.clear();
+    window.sessionStorage.clear();
+    const cookies = new Cookies();
+    for (const cookieName of Object.keys(cookies.getAll())) {
+      cookies.remove(cookieName);
+    }
+
+    // Reload the page. Seems to be the only way to stop analytics.
+    window.location.href = '/cookies';
```
craiga: Sometimes, the cookie page will be loaded with the URL `/cookies/` instead of `/cookies`. I don't know why. When this happens, reloading the page using `window.location.refresh()` or `window.history.go(0)` will cause a 404.

pataruco:
[This can help](https://github.com/zeit/next.js/issues/6080#issuecomment-496820953) to resolve the error log on console about `window is not defined`

TimboTambo:
QA feedback:
- window undefined issue on server - need to check if defined (i.e. if you're on server or browser) and provide a dummy window object if not
- when a user enables cookies for Google Analytics, we are also enabling local/session storage to log whether they've acknowledged the disclaimer/accepted cookies but we don't mention this at all on the cookie policy page. Is that intended?
- can't access the Zeplin designs currently to QA the text matches
Other than that, LGTM

TimboTambo:
Re-QA review:
* The "Save your preferences" button should be a button instead of an input
* I feel like the "Save your preferences" should take the user back to the homepage, or otherwise indicate that the preferences were saved
Otherwise looks good.

PR #286, opened 2020-02-03
Fix spelling of install

craiga:
Just fixing the spelling of the word "install".


rjkerrison:
I fixed this in a drive-by in #265

PR #293, opened 2020-02-05
Remove Zeebe manifests

StuartHarris:
This PR was a Spike to test working with hosted Zeebe – Now it just removes the local Kubernetes multi-node config (which doesn't really work, e.g. with Istio ingress)


StuartHarris:
Actually going to re-open, as this deletes the k8s config for Zeebe. Since this PR was raised, we are starting out with the hosted Zeebe at Camunda cloud (as running Zeebe locally under Istio is problematic).

StuartHarris:
Ignore all the commits in this PR. All of which were related to spiking Zeebe and are largely irrelevant now. So now it just deletes the `kind` environment (local cluster) which doesn't work anyway.

PR #444, opened 2020-02-24
Redis state manager

### LINKED ISSUE #434: AUTOMATIC BATCH PROCESS - State manager responds to requests by providing state of file
### User want
As a user
I want to see up to date documents on the products website
So I can make informed decisions

- [ ] State manager returns a file processing status in response to an API request.
- [ ] ~It indicates an error if the file cannot be found in the k:v store.~ As part of the design, it is possible that the state manager is legitimately called before it has seen a valid key. In the case where a valid key is given but no value is found for that key, it will return _NotFound_.
- [ ] _Accepted_, _Done_, _Error_ are valid states for the job to be in.
- [ ] Deserialisation from Redis works

**Data acceptance criteria**

**Size**

**Value**

**Effort**

### Exit Criteria met 
- [x] Backlog
- [x] Discovery
- [x] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

rjkerrison:
### Redis State Manager

Closes #434.

What a nice way to manage the state, Mr Palpatine


medicines/doc-index-updater/src/main.rs
```diff
@@ -12,29 +14,40 @@ async fn main() -> Result<(), Box<dyn error::Error>> {
...
-        env::var("PORT").unwrap_or_else(|e| {
-            eprintln!("defaulting PORT to {} ({})", PORT, e);
-            PORT.to_string()
-        })
-    )
-    .parse::<SocketAddr>()?;
+    let addr = format!("0.0.0.0:{}", get_env_or_default("PORT", PORT.to_string()))
+        .parse::<SocketAddr>()?;
+
+    let redis_addr = get_env_or_default("REDIS_ADDR", "redis://127.0.0.1:6379/".to_string());
```
pataruco: is port `6379` is the default of Redis? If is not can we parametised it?

StuartHarris: Yep it's the default, for local use. The endpoint is already parameterized so I think this is OK

PR #458, opened 2020-02-28
return 404 if job is not found

StuartHarris:
Now that we have immediate consistency it makes more sense to return HTTP `404 Not Found`, when there is no job related to the UUID specified in the URL path

### DIFF
--- a/medicines/doc-index-updater/src/state_manager/mod.rs
+++ b/medicines/doc-index-updater/src/state_manager/mod.rs
@@ -48,7 +48,11 @@ pub fn get_job_status(
 }
 
 async fn get_status_handler(id: Uuid, mgr: StateManager) -> Result<Json, Rejection> {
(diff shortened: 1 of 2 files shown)


medicines/doc-index-updater/src/state_manager/mod.rs
```diff
@@ -48,7 +48,11 @@ pub fn get_job_status(
 }
 
 async fn get_status_handler(id: Uuid, mgr: StateManager) -> Result<Json, Rejection> {
-    Ok(warp::reply::json(&mgr.get_status(id).await?))
+    let response = mgr.get_status(id).await?;
+    match response.status {
+        JobStatus::NotFound => Err(warp::reject::not_found()),
```
rjkerrison: try https://docs.rs/warp/0.1.6/warp/reply/fn.with_status.html

PR #474, opened 2020-03-05
Health probes

StuartHarris:
Now that we have a `/healthz` endpoint it is possible to add liveness and readyness probes to our `doc-index-updater` deployment. In order to support this when mTLS is enabled, the Istio configuration has to specify that the sidecar injector should rewrite the pod's IPTables to support this and we need a destination rule. See https://istio.io/docs/ops/configuration/mesh/app-health-check/.

This PR makes the necessary changes


medicines/doc-index-updater/examples/manifests/base/deployment.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       containers:
         - name: doc-index-updater
           image: doc-index-updater
-          imagePullPolicy: IfNotPresent
+          imagePullPolicy: Always
```
rjkerrison: should stay `IfNotPresent`

PR #477, opened 2020-03-06
Create manager grabs messages from Service Bus

### LINKED ISSUE #421: AUTOMATIC BATCH PROCESS - Create service picks up message from subscription
### User want
As a user
I want to see up to date documents on the products website
So I can make informed decisions

When a "create file" message is published to service bus, the message makes it to a topic and subscription that the Create Service is watching so that it picks up the message.

**Data acceptance criteria**

**Size**
M

**Value**

**Effort**

### Exit Criteria met 
- [ ] Backlog
- [ ] Discovery
- [ ] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

rjkerrison:
# Grabbing create messages from the Service Bus queue

Creates a **Create Manager**.

Closes #421.

Adam and God having a rock, paper, scissors

### Acceptance Criteria

- [ ] When a "create file" message is published to service bus, the message is picked up by the Create Service
- [ ] Tests cover the whole journey from calling the document_manager to picking up the Create message

### Testing information

You'll need the environment variables. You should be able to run `cargo test -- --ignored` and see all the tests pass.

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


medicines/doc-index-updater/tests/delete_manager.rs
```diff
@@ -1,14 +1,14 @@
...
 use doc_index_updater::{delete_manager, service_bus_client::delete_factory};
-use tokio_test::block_on;
+use support::get_ok;
 
 #[test]
 #[ignore]
 fn delete_manager_works() {
-    let sent_message = "This is the test message";
-    let mut delete_client = block_on(delete_factory()).unwrap();
+    let sent_message = "This is the delete test message";
```
m-doughty: This makes me think we should enforce the shape of the message somehow instead of using the `Client` directly.

medicines/doc-index-updater/src/create_manager/mod.rs
```diff
@@ -0,0 +1,29 @@
...
+        delay_for(Duration::from_secs(10)).await;
+    }
+    // TODO: create file in blob storage
+    // TODO: Update index
+    // TODO: Notify state manager
+}
+
+pub async fn get_message(create_client: &mut Client) -> Result<String, AzureError> {
+    let message = create_client
+        .peek_lock(time::Duration::days(1), Some(time::Duration::seconds(1)))
```
TimboTambo: This lock timeout may need to be increased, depending on how long it takes to SFTP, update the index and upload to the blob. Guess we can start with this default and lift it if nothing makes it through the pipeline!

medicines/doc-index-updater/tests/create_manager.rs
```diff
@@ -0,0 +1,20 @@
...
+#[ignore]
+fn create_manager_works() {
+    let sent_message = "This is the create test message";
+    let mut create_client = get_ok(create_factory());
+    get_ok(create_client.send_event(sent_message, time::Duration::seconds(1)));
+
+    let mut received_message = get_ok(create_manager::get_message(&mut create_client));
+    while received_message != sent_message {
+        received_message = get_ok(create_manager::get_message(&mut create_client));
+    }
```
TimboTambo: Would be good to have a max number of attempts here or some kind of timeout - if this test breaks it's going to get stuck in a loop.

rjkerrison: it'll break when it runs out of messages to check

medicines/doc-index-updater/src/create_manager/mod.rs
rjkerrison: definitely worth looking at increasing this when we work on the processing parts; as for now, if this timeout is too high, repeated testing runs will fail as the messages won't be unlocked on the second run

medicines/doc-index-updater/tests/delete_manager.rs
rjkerrison: Agree there's room for improvement here. Right now I think they're adding value. We might always want tests which only test the `client` calls, in which case we probably could use a `get_test_delete_message`. When we pick up tickets into Discovery, we should add the expected integration test changes into the ACs, in which case the decision to drop less relevant tests can be made when those ACs are met.

PR #478, opened 2020-03-06
Raise error with actual description of what went wrong

craiga:
Raise an error with the message from service bus instead of just "Server Error".

giphy-7

### DIFF
--- a/medicines/doc-index-updater/src/state_manager/redis.rs
+++ b/medicines/doc-index-updater/src/state_manager/redis.rs
@@ -29,8 +29,9 @@ impl From<RedisError> for MyRedisError {
 }
 
 impl From<MyRedisError> for warp::Rejection {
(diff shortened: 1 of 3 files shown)


medicines/doc-index-updater/src/state_manager/redis.rs
```diff
@@ -30,7 +30,7 @@ impl From<RedisError> for MyRedisError {
 
 impl From<MyRedisError> for warp::Rejection {
     fn from(_e: MyRedisError) -> Self {
-        warp::reject::custom(MyRedisError::Other(anyhow!("Server Error")))
+        warp::reject::custom(MyRedisError::Other(anyhow!(_e)))
```
rjkerrison: maybe call `tracing::error` for that error here, but still return the Rejection as `Server Error` to avoid leaking error information to the end-user

PR #502, opened 2020-03-12
Support XML requests for Sentinel

### LINKED ISSUE #486: AUTO BATCH - Support XML Requests
Accenture require an XML interface to the `doc-index-updater` API, as the Sentinel system is unable to serialize or deserialize JSON.

**Acceptance Criteria**

- The API should check the `Content-Type` or `Accept` headers. If the header is `application/xml`, then we should deserialize body & serialize response assuming XML rather than JSON;
- This should be implemented so that it's easy to carry forward into future endpoints.

**Stages**

- [x] Backlog
- [x] Discovery
- [x] Dev
- [x] Review
- [ ] QA
- [ ] UAT

m-doughty:
# Allows XML requests as well as JSON, and provides XML responses

resolves #486 

### Acceptance Criteria

- [x] If the user sends a request with `Content-Type` set to `application/xml`, we should assume the body is XML;
- [x] If the user sends a request with `Accept` set to `application/xml`, we should send our reply in XML;
- [x] Otherwise, we should default to JSON.

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/doc-index-updater/src/document_manager/mod.rs
+++ b/medicines/doc-index-updater/src/document_manager/mod.rs
@@ -1,11 +1,14 @@
 use crate::{
-    models::{CreateMessage, DeleteMessage, Document, JobStatus},
+    models::{CreateMessage, DeleteMessage, Document, JobStatus, XMLDocument},
     service_bus_client::{create_factory, delete_factory},
     state_manager::{with_state, StateManager},
 };
 use time::Duration;
 use uuid::Uuid;
-use warp::{reply::Json, Filter, Rejection, Reply};
+use warp::{
+    reply::{Json, Reply as ReplyReply, Xml},
+    Filter, Rejection, Reply,
+};
 
 #[derive(Debug)]
 struct FailedToDispatchToQueue;
@@ -14,10 +17,27 @@ struct FailedToDeserialize;
 impl warp::reject::Reject for FailedToDispatchToQueue {}
 impl warp::reject::Reject for FailedToDeserialize {}
 
+enum ReplyType {
+    Json,
+    Xml,
+}
+
+fn choose_reply_content_type(accept_header: Option<String>) -> ReplyType {
+    match accept_header {
+        Some(v) => match v {
+            "application/xml" => ReplyType::Xml,
+            "text/xml" => ReplyType::Xml,
+            _ => ReplyType::Json,
+        },
+        _ => ReplyType::Json,
+    }
+}
+
 async fn del_document_handler(
     document_content_id: String,
+    accept_header: Option<String>,
     state_manager: StateManager,
-) -> Result<Json, Rejection> {
+) -> Result<ReplyReply, Rejection> {
     if let Ok(mut queue) = delete_factory().await {
         let id = Uuid::new_v4();
         let message = DeleteMessage {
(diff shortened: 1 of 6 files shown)


m-doughty:
Pushing what I have so far. I'm just trying to make the Handlers play nice because `warp::reply::xml` and `warp::reply::json` have different return types, but that can wait for tomorrow.

medicines/doc-index-updater/src/document_manager/mod.rs
```diff
@@ -14,10 +17,27 @@ struct FailedToDeserialize;
...
+        _ => ReplyType::Json,
+    }
+}
+
 async fn del_document_handler(
     document_content_id: String,
+    accept_header: Option<String>,
     state_manager: StateManager,
-) -> Result<Json, Rejection> {
+) -> Result<ReplyReply, Rejection> {
```
TimboTambo: What's the rationale behind the naming - such a good reply they named it twice?

medicines/doc-index-updater/src/document_manager/mod.rs
```diff
@@ -27,9 +47,14 @@ async fn del_document_handler(
         let duration = Duration::days(1);
 
         match queue.send(message, duration).await {
-            Ok(_) => Ok(warp::reply::json(
-                &state_manager.set_status(id, JobStatus::Accepted).await?,
-            )),
+            Ok(_) => match choose_reply_content_type(accept_header) {
+                ReplyType::Xml => Ok(warp::reply::xml(
+                    &state_manager.set_status(id, JobStatus::Accepted).await?,
```
rjkerrison: is it possible to return just the type of `state_manager.set_status(id, JobStatus::Accepted).await` here, and then have separate Xml/Json wrapper calling this?

medicines/doc-index-updater/src/document_manager/mod.rs
```diff
@@ -27,20 +33,36 @@ async fn del_document_handler(
...
     }
 }
 
+async fn del_document_xml_handler(
+    document_content_id: String,
+    state_manager: StateManager,
+) -> Result<Xml, Rejection> {
+    let r: XMLJobStatusResponse = del_document_handler(document_content_id, state_manager)
+        .await?
+        .into();
```
rjkerrison: @m-doughty I had a go at making this use an idiomatic `into()`  - I think it's neater than calling the custom `from_…` method

m-doughty: I like this.

medicines/doc-index-updater/src/document_manager/mod.rs
```diff
@@ -27,20 +33,36 @@ async fn del_document_handler(
...
-            )),
+            Ok(_) => Ok(state_manager.set_status(id, JobStatus::Accepted).await?),
             Err(_) => Err(warp::reject::custom(FailedToDispatchToQueue)),
         }
     } else {
         Err(warp::reject::custom(FailedToDispatchToQueue))
     }
 }
 
+async fn del_document_xml_handler(
```
majikandy: I'm assume `del` means `delete`?  shall we actually write `delete` or is there a decent reason to keep as `del`?

if changing to delete then the other areas below with del would need changing too but I won't pop a comment there

medicines/doc-index-updater/src/models.rs
```diff
@@ -61,6 +61,7 @@ pub struct Document {
 
 #[derive(Serialize, Deserialize, Debug, PartialEq, Clone)]
 pub enum FileSource {
+    #[serde(alias = "sentinel")]
```
rjkerrison: https://docs.rs/serde-aux/0.6.1/serde_aux/container_attributes/fn.deserialize_struct_case_insensitive.html
might be able to use this for full case-insensitivity but for now title-case and lowercase covers what we've been doing 👍

medicines/doc-index-updater/src/document_manager/mod.rs
rjkerrison: agree: `delete` and `create` instead of `del` and `check_in` would align with the names for `*_manager` and `*Message` that we've got elsewhere

Issue #511, opened 2020-03-13
Searching for a term with "%" in throws error


TimboTambo:
**Describe the bug**
Searching for a term with "%" in throws error and causes next.js to crash.

**To Reproduce**
1) Search for `bacteriostatic solvent (0.3%w/v metacresol in wfi)` 
2) Observe error page

**Expected behavior**
The search results page should load, at the very least, and should return results if there are any.

**Screenshots**
<img width="766" alt="Screen Shot 2020-03-13 at 17 31 49" src="https://user-images.githubusercontent.com/6436551/76645347-bc599d00-6550-11ea-9694-e46b48c8e6c8.png">

**Additional context**
An unnecessary `decodeURIComponent` is being performed on the query, as it has already been decoded by that point, so when it finds a standalone `%` it gets confused.

chel-mhra:
Can you get this one over to Andy N to help QA please?

chel-mhra:
@TimboTambo @ainemckay Can you get this one over to Andy N to help QA please?

TimboTambo:
@chel-mhra sure! This wasn't included in the CR for this week so it'll only go out next week. It'll first be QA'd by someone in our team and then we'll identify with Andy N who would be best placed to test it from the business prior to next week's CAB.

craiga:
✅ Passed QA

PR #514, opened 2020-03-14
Timlee/fix search for percent

TimboTambo:
# Fix search for `%`
See details in bug #511 

### Acceptance Criteria

- [ ] Searching for terms with `%` in should not throw an error

### Testing information
Search for a term with `%` in - no error should be thrown. E.g. `bacteriostatic solvent (0.3%w/v metacresol in wfi)`

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/web/src/services/search-query-normalizer.test.ts
+++ b/medicines/web/src/services/search-query-normalizer.test.ts
@@ -7,8 +7,12 @@ describe(buildFuzzyQuery, () => {
   });
 
   it('ignores special characters', () => {
-    const fuzzyQuery = buildFuzzyQuery('hello*');
-    expect(fuzzyQuery).toBe('hello~1 hello^4');
+    const fuzzyQuery = buildFuzzyQuery(
+      'bacteriostatic solvent (0.3%w/v metacresol in wfi)',
+    );
+    expect(fuzzyQuery).toBe(
+      'bacteriostatic~1 bacteriostatic^4 solvent~1 solvent^4 0.3~1 0.3^4 w~1 w^4 v~1 v^4 metacresol~1 metacresol^4 in~1 in^4 wfi~1 wfi^4',
+    );
   });
 
   it('builds fuzzy words from product name', () => {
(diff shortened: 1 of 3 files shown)


medicines/web/src/services/search-query-normalizer.test.ts
```diff
@@ -7,8 +7,12 @@ describe(buildFuzzyQuery, () => {
   });
 
   it('ignores special characters', () => {
-    const fuzzyQuery = buildFuzzyQuery('hello*');
-    expect(fuzzyQuery).toBe('hello~1 hello^4');
+    const fuzzyQuery = buildFuzzyQuery(
+      'bacteriostatic solvent (0.3%w/v metacresol in wfi)',
```
craiga: :+1:

medicines/web/src/services/search-query-normalizer.ts
```diff
@@ -28,7 +28,7 @@ const addNormalizedProductLicenses = (q: string): string => {
 };
 
 const splitByNonSearchableCharacters = (query: string) =>
-  query.split(/(?:[,+\-!(){}\[\]^~*?:\/]|\s+)/gi);
+  query.split(/(?:[,+\-!(){}\[\]^~*?:%\/]|\s+)/gi);
```
craiga: Is there not a way we can encode these characters for Azure?

If people are really searching for "100%awesome thing", then we probably don't want to.

TimboTambo: The current approach is to remove these symbols by splitting on them. I played around with encoding them instead and it seemed tricky to do this whilst constructing our fuzzy search from the broken up search terms. I wanted to get this quick fix in, but I agree it would be better to preserve the special characters used in the search by encoding them, instead of clearing them out. I'll create a card for this.

Issue #498, opened 2020-03-12
Some cookies still present after clear


TimboTambo:
**Describe the bug**
After accepting cookies and then turning them off, some cookies are still present (`_ga` and `_gat`).

**To Reproduce**
1) Accept all cookies. 
2) Go to cookie policy page and disable cookies then save
3) Examine cookies in developer tools
4) Navigate to another page
5) See `_ga` and `_gat` cookies get added back 

**Expected behavior**
They should clear and remain cleared once cookies are set to disabled.

**Screenshots**
N/A

**Additional context**
Not sure if this has ever worked fully as intended.

craiga:
In QA, I found that cookies are being cleared effectively, but when the user clicks "Accept all cookies":
 * cookies aren't being set; and
 * Google Analytics isn't being activated.

In order for both of these to happen.

I think we need to trigger a page reload after allowing cookies for Google Analytics to do its initialisation.

m-doughty:
I'm comfortable saying that this works as intended in Brave, but I can't test in Safari.

majikandy:
I can confirm this works on Dev environment in both `Safari` and `Chrome` on a mac.
<img width="1385" alt="Screenshot 2020-03-24 at 11 31 28" src="https://user-images.githubusercontent.com/444711/77421062-106f4780-6dc3-11ea-99d0-4d450ee5b2c3.png">

moving forward to ready for UAT

AndySnowMHRA:
Hi there still appear to be some analytics in Edge browser (unless I am reading it wrong. working fine in Chrome.

PR #515, opened 2020-03-16
Clear cookies by setting path and domain

TimboTambo:
# Clear all cookies

Relates to bug #498

### Acceptance Criteria

- [ ] All cookies are cleared

### Testing information

1) Accept cookies. 
2) Go to cookies policy page and disallow cookies
3) Refresh page
4) Check cookies in dev tools - they should be empty

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


medicines/web/src/components/cookie-form/index.tsx
```diff
@@ -57,6 +57,8 @@ const StyledContentWrapper = styled.main`
   }
 `;
 
+const domain = '.windows.net';
```
craiga: Should be `.mhra.gov.au`, or derived from `window.location`.

TimboTambo: The domain of the GA cookies specifically is `.windows.net` and require this specifically to be cleared, so you can't auto extract this from the window location and you also can't directly access the cookie domain (or other metadata) from JS, I believe. This could be moved into an environment variable, which I see as only being necessary if the site was going to be deployed on a different platform.

craiga: (lol @ me for the `.au` above; old habits die hard)

The domain of the cookies in production is `.mhra.gov.uk`. As far as I understand it, attempting to set a cookie on a different domain just won't work.

Screenshot 2020-03-16 at 11 48 40

TimboTambo: Actually, I was looking at the staging site, so ignore this. I'll have a think about the best way to do this.

medicines/web/src/components/cookie-form/index.tsx
```diff
@@ -57,6 +57,8 @@ const StyledContentWrapper = styled.main`
   }
 `;
 
+const cookieDomain = process.env.ROOT_URL_DOMAIN as string;
```
craiga: I don't suppose there's any way to default this to `localhost` so it doesn't need to be set when running locally?

PR #233, opened 2019-12-19
CDN infrastructure

callerc1:
### Acceptance Criteria

For both environments (pros/non-prod)
- [x] Add CDN profile
- [x] Add Producst endpoint 
- [x] Add CPD endpoint

### Testing information

<insert notes for testers that might be helpful>

### Pre-flight Checklist

- [ ] Tests
- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

PR #526, opened 2020-03-18
move cluster to the correct subnet

StuartHarris:
Moves the `non-prod` cluster into the correct subnet. This also entails moving to `uksouth` (from `westeurope`)


.github/workflows/doc-index-updater.yaml
```diff
@@ -19,23 +19,23 @@ jobs:
...
-        working-directory: ./products/medicines/doc-index-updater
-        run: |
-          TAG="$(git rev-parse --short=7 ${{ github.sha }})"
-          make ci-branch tag=$TAG
+      # - name: Build and test Docker image (BRANCH only)
+      #   if: github.ref != 'refs/heads/master'
+      #   working-directory: ./products/medicines/doc-index-updater
+      #   run: |
+      #     TAG="$(git rev-parse --short=7 ${{ github.sha }})"
+      #     make ci-branch tag=$TAG
```
pataruco: I think you left this comment to speed up your feedback loop

StuartHarris:
Don't merge this yet. I removed the `development` and `prod` environments because they were getting out of date and we really only have a `non-prod` cluster. However I forgot about the CDN configuration in #233 which we do need. I think we have to bring that PR in first, and fix the `prod` environment so that it just has CDN (and other relevant) stuff.

StuartHarris:
I reinstated the prod env (without a cluster) so that we don't lose the current production configuration and we will be able to merge #233 to master. So I think this can be reviewed now

infrastructure/environments/prod/main.tf
```diff
@@ -44,17 +40,6 @@ module cpd {
 
   environment         = var.ENVIRONMENT
   location            = var.REGION
+  namespace           = local.namespace
   resource_group_name = azurerm_resource_group.products.name
 }
-
-# Service Bus
-module service_bus {
```
sgrowe: We don't want a service bus or cluster in prod?

StuartHarris:
@sgrowe We do indeed need service bus in production, but I wanted to remove all the resources that we haven't built yet in production, as non-prod was changing and keeping them both in sync is currently unneccessary. I think it will be easier to update prod config later when it's settled down.

infrastructure/environments/prod/variables.tf
```diff
@@ -8,15 +8,7 @@ variable "RESOURCE_GROUP_PRODUCTS" {
...
-}
-
-variable "CLIENT_SECRET" {
-  description = "Service Principal Client Secret"
-}
-
 variable "ENVIRONMENT" {
   description = "Environment name"
-  default     = "prod"
+  default     = "non-prod"
```
sgrowe: How come you've renamed this to non-prod? 🤔 

If the clusters are now named after `var.environment` won't that cause a clash if we try to spin them both up?

StuartHarris: Thanks, fixed that now

Issue #422, opened 2020-02-17
AUTOMATIC BATCH PROCESS - Delete service informs State Manager of completion / error


ainemckay:
### User want
As a user
I want to see up to date documents on the products website
So I can make informed decisions

**Customer acceptance criteria** 

**Technical acceptance criteria**
Delete service calls sTATE mANAGER with the status of the delete.

**Data acceptance criteria**

**Testing acceptance criteria** 

**Size**
S

**Value**

**Effort**

### Exit Criteria met 
- [ ] Backlog
- [ ] Discovery
- [ ] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

majikandy:
QA
Status comes back on the job as `Accepted` 
When service worker picks it up it shows the error on standard out

But... the job status remains as `Accepted`

```
{
   "id": "d69aba65-00cf-454f-9665-ff63f0cefab2",
   "status": "Accepted"
}
```

majikandy:
QA'd the whole of delete again.

Can confirm it now works as expected 👍 

```
{
    "id": "3c1cb9a5-370d-4c72-99e8-39c44db944ae",
    "status": {
        "Error": {
           "message": "Cannot find document with content ID CON1579848082866",
           "code": ""
        }
     }
}
```

PR #527, opened 2020-03-18
Handle errors in delete worker

craiga:
* Set error state in state manager.
* Remove message from queue.
* Reflect the structure of the create worker in the delete worker.

Resolves issues found during testing of #422.

giphy-16

### DIFF
--- a/medicines/doc-index-updater/src/create_manager/mod.rs
+++ b/medicines/doc-index-updater/src/create_manager/mod.rs
@@ -56,13 +56,15 @@ async fn try_process_from_queue(
                     == "Couldn't retrieve file: [-31] Failed opening remote file".to_string()
                 {
                     tracing::info!("Updating state to errored and removing message");
-                    let _ = state_manager.set_status(
-                        retrieval.message.job_id,
-                        JobStatus::Error {
-                            message: "Couldn't find file".to_string(),
-                            code: "404".to_string(),
-                        },
-                    );
+                    let _ = state_manager
+                        .set_status(
+                            retrieval.message.job_id,
+                            JobStatus::Error {
+                                message: "Couldn't find file".to_string(),
+                                code: "404".to_string(),
+                            },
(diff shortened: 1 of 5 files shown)


medicines/doc-index-updater/src/create_manager/mod.rs
```diff
@@ -56,13 +56,15 @@ async fn try_process_from_queue(
...
-                    );
+                    let _ = state_manager
+                        .set_status(
+                            retrieval.message.job_id,
+                            JobStatus::Error {
+                                message: "Couldn't find file".to_string(),
+                                code: "404".to_string(),
+                            },
+                        )
+                        .await?;
```
craiga: I found that `.await?` was required to set the status in the delete process.

TimboTambo: Ah yeah - noticed this yesterday when I was looking into and completely forgot to do anything with it!

medicines/doc-index-updater/src/models.rs
```diff
@@ -29,11 +30,18 @@ impl FromStr for JobStatus {
...
-                code: "0x0".to_owned(),
-            }),
-            e => Err(format!("Status unknown: {}", e)),
+            status => {
+                // If this message is in the format "Error(error code: error message)",
+                // reconstruct it into JobStatus::Error.
+                let error_re = Regex::new(r"^Error\((?P<code>[^:]*): (?P<message>.*)\)$").unwrap();
+                match error_re.captures(status) {
+                    Some(capture) => Ok(JobStatus::Error {
+                        message: capture.name("message").unwrap().as_str().to_string(),
```
TimboTambo: These go through quite a journey! A result -> str -> String!

craiga: It's nuts. The captures have an `as_str` but no `to_string`.

PR #539, opened 2020-03-20
Toggle JSON logs based on environment variable

sgrowe:
You can now set the enviroment variable `JSON_LOGS=false` to see the pretty logs we used to have


m-doughty:
I'm really interested to know what magic is happening behind the scenes to coerce `"false"` to `false`. I'd be tempted to refactor this as `LOG_FORMAT` and default to `"json"`, it stops weird undefined behaviors like `JSON_LOGS=xyz cargo run` and allows for more extensibility going forward.

sgrowe:
> I'm really interested to know what magic is happening behind the scenes to coerce `"false"` to `false`. I'd be tempted to refactor this as `LOG_FORMAT` and default to `"json"`, it stops weird undefined behaviors like `JSON_LOGS=xyz cargo run` and allows for more extensibility going forward.

`get_env_or_default` is a helper that was already in the code base [that calls `.parse()`](https://github.com/MHRA/products/blob/562d4589a311129f9a6e52d1872ed8560164e698/medicines/doc-index-updater/src/main.rs#L84) on the environment variable for you, so it's just Rust standard parsing. And if it fails to parse it just defaults to `true` (and prints a warning)

majikandy:
The main reason for the decision on reason on behind JSON_LOGS=false/true and defaulting to true is simply so that there is an override for devs locally if you know about it.

so JSON_LOGS=xyz giving json logs is fine.

the only downside being JSON_LOGS=NO and it still does json, but I don’t think any developer would complain and this feature would only be for developers I think.

medicines/doc-index-updater/src/main.rs
```diff
@@ -10,15 +10,12 @@ const PORT: u16 = 8000;
...
-        .json()
-        .flatten_event(true)
-        .with_timer(tracing_subscriber::fmt::time::ChronoUtc::rfc3339())
-        .with_max_level(Level::INFO)
-        .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
-        .finish();
+    if get_env_or_default("JSON_LOGS", true) {
+        use_json_log_subscriber()
+    } else {
+        use_unstructured_log_subscriber()
```
TimboTambo: Small suggestion - might make unit testing easier (when we get around to that) if these functions returned a subscriber that you then `set_global_default` with on the next line.

majikandy: the only reason we didn't do that is because the .json function then changes the generic type slightly

medicines/doc-index-updater/.env.example
```diff
@@ -22,3 +22,5 @@ SENTINEL_SFTP_USERNAME=
 SERVICE_BUS_NAMESPACE=
 
 STORAGE_CONTAINER=
+
+JSON_LOGS=
```
majikandy: cheers @rjkerrison

Issue #427, opened 2020-02-18
AUTOMATIC BATCH PROCESS - Create service uploads file to blob container


KaylieGreen1607:
### User want
As a user
I want to see up to date documents on the products website
So I can make informed decisions

**Customer acceptance criteria** 

**Technical acceptance criteria**
Create service uploads a file to the blob container. A lot of this code can be pulled from the current import process.

**Data acceptance criteria**

**Testing acceptance criteria** 

**Size**

**Value**

**Effort**

### Exit Criteria met 
- [x] Backlog
- [x] Discovery
- [x] DUXD
- [x] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

craiga:
It looks like the create worker will panic when it can't connect to the SFTP server.

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 61, kind: ConnectionRefused, message: "Connection refused" }', src/create_manager/sftp_client.rs:10:15
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

After this, the create worker ceases to respond to messages.

To reproduce, fire up the create worker without a running SFTP server.

PR #548, opened 2020-03-23
Helpful bad STORAGE_MASTER_KEY error

rjkerrison:
@craiga discovered while testing #427 that if his `STORAGE_MASTER_KEY` is weird, i.e. not a valid base64-encoded string, he gets an unhelpful and confusing error message about an unwrap at `azure_sdk_storage_core/src/rest_client.rs:415:17`.

This PR is to ensure we do a minimal check to highlight this problem, so devs spend less time hunting through dependent crates when their setup is marginally wrong.

### Acceptance Criteria

- [ ] When the retrieved `STORAGE_MASTER_KEY` is not a valid base64-encoded value, throw an error there and then

### Testing information

Add a character at the start of a valid `STORAGE_MASTER_KEY` to make it the wrong length.

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/doc-index-updater/Cargo.toml
+++ b/medicines/doc-index-updater/Cargo.toml
@@ -14,7 +14,7 @@ azure_sdk_service_bus = { git = "https://github.com/redbadger/AzureSDKForRust", 
 azure_sdk_storage_blob = { git = "https://github.com/redbadger/AzureSDKForRust", branch = "service-bus-peek-lock-timeout" }
 azure_sdk_storage_core = { git = "https://github.com/redbadger/AzureSDKForRust", branch = "service-bus-peek-lock-timeout" }
 chrono = "0.4.11"
-base64 = "0.12.0"
+base64 = "0.11.0"
 futures = "0.3.4"
 hyper = "0.13"
 lazy_static = "1.4.0"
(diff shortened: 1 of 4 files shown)


medicines/doc-index-updater/Cargo.toml
```diff
@@ -14,7 +14,7 @@ azure_sdk_service_bus = { git = "https://github.com/redbadger/AzureSDKForRust",
 azure_sdk_storage_blob = { git = "https://github.com/redbadger/AzureSDKForRust", branch = "service-bus-peek-lock-timeout" }
 azure_sdk_storage_core = { git = "https://github.com/redbadger/AzureSDKForRust", branch = "service-bus-peek-lock-timeout" }
 chrono = "0.4.11"
-base64 = "0.12.0"
+base64 = "0.11.0"
```
craiga: Should we be moving backwards here?

rjkerrison: Unfortunately it was the choice between going back one version here, or updating the Azure SDK crate we're using. We've forked and branched that crate already, but we're close to getting that PR merged. Our use of this crate is small (one line outside this check) so it didn't seem worth upping that package for no real benefit.

craiga: Fair enough.

PR #560, opened 2020-03-25
Update architecture diagram

StuartHarris:
Update architecture diagram


TimboTambo:
As we're not using CosmosDB anymore, can you pull this from the diagram? Also it may be worth (now or later) tweaking the description as some bits are quite out of date, e.g. the importer isn't being run on a scheduler.

StuartHarris:
@TimboTambo Thanks, updated the readme (should be a bit better now) and the diagram

PR #571, opened 2020-03-26
Add dev environment for testing infra changes

sgrowe:
Adds a dev environment to MHRA Azure.

⚠️ ⚠️ 
It currently uses the CIDR block that is reserved for prod so as to avoid clashing with any other virtual networks MHRA may have. Obviously we should change this as soon as possible


infrastructure/environments/dev/outputs.tf
```diff
@@ -0,0 +1,65 @@
...
+  # TODO: fragile
+  value = local.namespace
+}
+
+output "storage_master_key" {
+  value = module.products.storage_access_key
+}
+
+
+# output "cpd_primary_access_key" {
```
rjkerrison, line 59: do we need these comments?

infrastructure/environments/non-prod/main.tf
```diff
@@ -19,6 +20,11 @@ terraform {
   }
 }
 
+locals {
+  namespace        = "mhraproductsnonprod"
```
rjkerrison, line 24: should this use `var.ENVIRONMENT` too?

infrastructure/modules/cluster/main.tf
```diff
@@ -16,41 +16,23 @@ resource "azurerm_virtual_network" "cluster" {
   address_space       = [var.vnet_cidr]
 }
 
-data "azurerm_route_table" "load_balancer" {
```
rjkerrison, line 19: how come we're removing the load_balancer in this PR? description suggests it's just about creating a dev environment

infrastructure/environments/non-prod/main.tf
sgrowe: Yeah, I think ideally it would be but `var.ENVIRONMENT` was already set to `non-prod` with a dash and the namespace without a dash and I don't want to change the names of things in non-prod and force them to be recreated so I've just stuck to the naming that's already there for the time being

infrastructure/modules/cluster/main.tf
sgrowe: So this is the routing table for the load balancer which MHRA created for us. Here we're setting it up as a data source so that we can read info about its state. The dev environment needs to create its own routing table rather than read from one that already exists so I've moved this out of the `cluster` module into [`non-prod/main.tf`](https://github.com/MHRA/products/pull/571/files#diff-1da41d11e3dc7d7c623bc04b68d99563) (and `dev/main.tf`) and then just pass in the attributes that are needed as variables.

PR #576, opened 2020-03-27
Run tests on feature branches with slashes in

TimboTambo:
# Run tests on feature branches with slashes in

Current Github actions use `"*"` to identify all branch names, but to also match branches with slashes in (e.g. `feature/branch`) the pattern should use `"*/*"` as detailed here: https://github.community/t5/GitHub-Actions/Github-Actions-on-push-not-master/m-p/31459/highlight/true#M796

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/.github/workflows/doc-index-updater-branch.yaml
+++ b/.github/workflows/doc-index-updater-branch.yaml
@@ -3,7 +3,7 @@ name: doc-index-updater-branch
 on:
   push:
     branches:
-      - "*"
+      - "*/*"
       - "!master"
     paths:
       - medicines/doc-index-updater/**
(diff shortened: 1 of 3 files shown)


.github/workflows/doc-index-updater-branch.yaml
```diff
@@ -3,7 +3,7 @@ name: doc-index-updater-branch
 on:
   push:
     branches:
-      - "*"
+      - "*/*"
```
StuartHarris: OK, do we need "*" as well though (for branch names that _don't_ have a slash)

StuartHarris: Ahh I think it should actually be `**` (https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet)

TimboTambo: Ah yes! I was lead astray by one comment and didn't research it further :)

PR #587, opened 2020-03-30
Nifty delete testing script

majikandy:
# Delete shell script, runs status check after 

I made this script when testing DELETE but figured it may be useful for others and could even be the start of a "the system hangs together" end to end test

### Acceptance Criteria

- [ ] Runs the curl DELETE command to delete the specified file 
- [ ] polls the status service for that job and prints the output

### Testing information

`sh delete.sh CON_ID`

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


majikandy:
@rjkerrison Yes I was in 2 minds about whether to have it in tests for it to evolve into a test - moved to scripts for now - also threw in the check for it being deleted from blob storage too in that last push

StuartHarris:
This is great! One thing that might be useful though, is to use `shellcheck` and `shellfmt` etensions in `vscode` (I find them invaluable for finding bugs etc and tidying up shell scripts). In order for the `shellcheck` extension to work you also need to `brew install shellcheck`.
They are invaluable though (my vscode shows 20 problems in this script!).

majikandy:
Thanks for this - I normally bash them out in vi but nice tip, I always fine shell scripts a bit all over the place and would be nice to be more idiomatic.  I’ll run that over it now

PR #599, opened 2020-04-01
Let created and release_state be null.

craiga:
Let the fields `created` and `release_state` be null.

giphy-20

### DIFF
--- a/medicines/doc-index-updater/stunnel.conf
+++ b/medicines/doc-index-updater/stunnel.conf
@@ -4,4 +4,4 @@ foreground = yes
 [doc_index_updater]
 client = yes
 accept = 127.0.0.1:6379
-connect = doc-index-updater-non-prod.redis.cache.windows.net:6380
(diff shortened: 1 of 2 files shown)


medicines/doc-index-updater/stunnel.conf
```diff
@@ -4,4 +4,4 @@ foreground = yes
 [doc_index_updater]
 client = yes
 accept = 127.0.0.1:6379
-connect = doc-index-updater-non-prod.redis.cache.windows.net:6380
+connect = doc-index-updater-dev.redis.cache.windows.net:6380
```
rjkerrison, line 7: did you mean to commit this? I want it in master anyway because I've had to do it locally too, so I don't mind

or… we could get this to read from `.env`?

craiga: Can we? We don't have any code here AFAIK.

PR #600, opened 2020-04-01
Fix the state race condition

### LINKED ISSUE #598: Status on accepting a job is updated late
Context: doc-index-updater, `/medicines/doc-index-updater`.

Currently, a job state is updated to _Accepted_ after adding to the service bus.

This could be causing an issue if the job is picked up and fully processed before the state is updated. The State would then be updated from _Done_ or _Error_ to _Accepted_, and it would remain in the _Accepted_ state indefinitely.

Instead, we should update to _Accepted_ before pushing to the service bus. We should then update to _Error_ only if there's a problem pushing to the service bus.

## To Reproduce

Steps to reproduce the behaviour:

1. Set `SECONDS_TO_WAIT` to `0`.
2. Set the `STORAGE_MASTER_KEY` to something which isn't base64-encoded.
3. Add a delete job
4. If the job is picked up quickly enough, the job will immediately error but the last state will be _Accepted_, not _Error_.
5. Check the job status.

rjkerrison:
# Accept first, ask questions later

Fixes #598.

Sputnik

### Acceptance Criteria

- [x] Set status to _Accepted_ before pushing to queue
- [ ] Set status to _Error_ if there's an issue pushing to queue

### Testing information

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/doc-index-updater/src/document_manager/mod.rs
+++ b/medicines/doc-index-updater/src/document_manager/mod.rs
@@ -1,10 +1,10 @@
 use crate::{
     auth_manager,
     models::{
-        CreateMessage, DeleteMessage, Document, JobStatus, JobStatusResponse, XMLDocument,
+        CreateMessage, DeleteMessage, Document, JobStatus, JobStatusResponse, Message, XMLDocument,
         XMLJobStatusResponse,
     },
-    service_bus_client::{create_factory, delete_factory},
+    service_bus_client::{create_factory, delete_factory, DocIndexUpdaterQueue},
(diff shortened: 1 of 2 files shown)


rjkerrison:
Yeah, I checked the scenario where you get something other than "error" and it's easily reproducible using the steps to reproduce, especially now you can just sabotage your queue or stay offline with a local redis. If we can process the job that quickly, let's report it that quickly.

PR #622, opened 2020-04-06
Error log message that was missing

majikandy:
If the queue doesn't exist then we don't get an error message when a request comes in.

### Acceptance Criteria

- [ ] Error in the logs when queue not found

### Testing information
Point at a non-existent queue and try a delete or create

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/doc-index-updater/src/document_manager/mod.rs
+++ b/medicines/doc-index-updater/src/document_manager/mod.rs
@@ -56,6 +56,7 @@ where
                     },
                 )
                 .await?;
+            tracing::error!("Failed to dispatch to queue. Check API keys");
             Ok(state)
         }
     }
--- a/medicines/doc-index-updater/src/document_manager/mod.rs
+++ b/medicines/doc-index-updater/src/document_manager/mod.rs
@@ -48,7 +48,7 @@ where
         Ok(_) => Ok(state_manager.get_status(message.get_id()).await?),
         Err(e) => {
             tracing::error!(
-                "{:?}. Failed to dispatch to queue. Check API Keys or Policy values in environment variables.",
+                "Failed to dispatch to queue. Check environment variables align for queue names, policies and keys. Error: ({:?})",
                 e
             );
             let state = state_manager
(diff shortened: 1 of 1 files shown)


medicines/doc-index-updater/src/document_manager/mod.rs
```diff
@@ -54,6 +54,7 @@ where
                     },
                 )
                 .await?;
+            tracing::error!("Failed to dispatch to queue. Check API keys");
```
rjkerrison, line 57: can you put this above the state_manager call? a failure in state_manager will prevent us from hitting this line. You could also use `Err(e)` and print the true error message of `e`.

majikandy:
@craiga @TimboTambo @rjkerrison 

Added in the instrumentation to get correlation id in it too

log message now looks like this:-

```
ERROR document_manager{correlation_id="3c9ab764-4f84-4f45-85ac-a9a6678d17cf"}: doc_index_updater::document_manager: UnexpectedHTTPResult(UnexpectedHTTPResult { expected: [201], received: 401, body: "" }). Failed to dispatch to queue. Check API Keys or Policy values in environment variables.
```

medicines/doc-index-updater/src/document_manager/mod.rs
```diff
@@ -67,12 +72,16 @@ async fn delete_document_handler(
...
+        let correlation_id = id.to_string();
+        let correlation_id = correlation_id.as_str();
 
         let message = DeleteMessage {
             job_id: id,
             document_content_id,
         };
-        queue_job(&mut queue, &state_manager, message).await
+        queue_job(&mut queue, &state_manager, message)
+            .instrument(tracing::info_span!("document_manager", correlation_id))
```
craiga, line 83: :+1:

Would be nice to do the same in create document handler.

majikandy: true, this is like never ending story :)

majikandy: Done

PR #624, opened 2020-04-06
sftp: replace assert panic with error

StuartHarris:
Replaces an `assert!` panic with a returned error if sftp session is not authenticated.

### DIFF
--- a/medicines/doc-index-updater/src/create_manager/sftp_client.rs
+++ b/medicines/doc-index-updater/src/create_manager/sftp_client.rs
@@ -1,30 +1,17 @@
 use crate::models::FileSource;
+use anyhow::anyhow;
 use ssh2::{File, Session, Sftp};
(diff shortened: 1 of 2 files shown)


medicines/doc-index-updater/src/create_manager/sftp_client.rs
```diff
@@ -1,30 +1,17 @@
...
+use anyhow::anyhow;
 use ssh2::{File, Session, Sftp};
-use std::{fmt::Display, io::Read, net::TcpStream};
+use std::{io::Read, net::TcpStream};
+use thiserror::Error;
 
-#[derive(Debug)]
+#[derive(Error, Debug)]
 enum SentinelSftpError {
+    #[error("A TCP error connecting to server. ({0:?})")]
```
rjkerrison, line 9: nice, I did not know this shorthand

PR #625, opened 2020-04-06
Log error when reading queue fails

majikandy:
### Acceptance Criteria

- [ ] Logs an error when can't read from queue 

### Testing information

set env variables wrong and run `make`


majikandy:
Note: this also fixes the incorrect message we get when the Keys/Queue are wrong. 
```
ERROR doc_index_updater::service_bus_client: Message found could not be parsed (""). Deleting it.
```

PR #626, opened 2020-04-07
Add script for populating medicines/search/.env

sgrowe:
Saves a bit of time updating `medicines/search/.env` every day

### DIFF
--- a/infrastructure/scripts/update-search-env.sh
+++ b/infrastructure/scripts/update-search-env.sh
@@ -1,0 +1,52 @@
+#!/bin/bash
+set -euo pipefail
+
+# Updates the .env file at medicines/search/.env to have the the correct environment variables
+
+# Prerequisites:
+# - `jq` needs to be installed - https://stedolan.github.io/jq/
+
+PS3='Please enter your current environment: '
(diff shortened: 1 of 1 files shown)


infrastructure/scripts/update-search-env.sh
```diff
@@ -0,0 +1,52 @@
...
+    echo "STORAGE_ACCOUNT=$(echo "$OUTPUT" | jq .storage_account_name.value --raw-output)"
+    echo "STORAGE_CONTAINER=docs"
+    echo "STORAGE_MASTER_KEY=$(echo "$OUTPUT" | jq .storage_master_key.value --raw-output)"
+} >../../medicines/search/.env
+
+echo
+echo "Updated medicines/search/.env"
+
+echo
+echo "Don't forget to run make set-env to save the new variables!"
```
rjkerrison, line 52: maybe here we could do
```
cd ../../medicines/search
make set-env-dev
```
?

sgrowe: Yeah, good shout. I want to just try it out a couple more times just to make sure it works properly first but yeah I'll add that at a later date

PR #630, opened 2020-04-07
Azure Monitor dashboard

### LINKED ISSUE #525: Observability | Azure monitor dashboard
## User want

As a technical user 
I want to visualize doc index updater using Azure monitor
So I can monitor and alert

## Technical acceptance criteria

Azure Monitor should have the tooling to visualise the [golden signals](https://landing.google.com/sre/sre-book/chapters/monitoring-distributed-systems/#xref_monitoring_golden-signals):

- [x] Latency
- [x] Traffic
- [x] Errors
- [x] Saturation

The four golden signals of monitoring are latency, traffic, errors, and saturation. If you can only measure four metrics of your user-facing system, focus on these four.

### Latency

Latency corresponds to the time it takes to serve a request.

We should separate success(200~300) and error latencies(400~500).

#### Way to measure it:

Store latency metrics as histogram time-series. We put our metrics into buckets that our exporters collect every minute. This way we can calculate the n-quantiles for our services’ latencies.

For `0 < n < 1` and a histogram with q values, the n-quantile of that histogram is the value that ranks `n*q` amongst the q values. This means the 0.5-quantile (aka. the median) for a histogram with x entries is the value for which half of the x values are smaller or equal.

Now, because we are using Prometheus, we must be very careful with the bucket sizes we pick. Prometheus allows for both linear and exponential bucket sizes. It doesn’t really matter which of the two options you choose. As long as you factor [estimation errors](https://prometheus.io/docs/practices/histograms/#errors-of-quantile-estimation) in your choice, you’re good to go.

Prometheus doesn’t give the exact value for a quantile. It actually detects which of the buckets that quantile is in. Once it has done that, it uses linear interpolation to give an estimated value.

### Traffic

Count how many requests it receives every second. Now, because we only get new metrics every minute, we can’t get an exact value for a given second. We can use [Prometheus’ rate and irate](https://www.robustperception.io/irate-graphs-are-better-graphs) functions to work with averaged numbers of queries per second instead.

### Errors

The explicit error rate is pretty easy to compute, as we only need to divide HTTP **500**s served by the total number of requests served. Like for traffic, we use an averaged value.

### Saturation

To watch a service’s saturation, we need to determine its constraining parameters. We can measuring both CPU and Memory usage Kubernetes internals and [`kube-state-metrics`](https://github.com/kubernetes/kube-state-metrics) allow us to get such metrics for containers.

> Snipets taken from [Distributed Monitoring 101: the “Four Golden Signals”](https://medium.com/forepaas/distributed-monitoring-101-the-four-golden-signals-305bbbc33d35)

**Size**

**Value**

**Effort**

### Exit Criteria met 
- [ ] Backlog
- [ ] Discovery
- [ ] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

sgrowe:
Use [metrics pulled from prometheus](https://github.com/MHRA/deployments/pull/12) to show the status of the four "golden signals" listed in #525.

Current charts:
- CPU usage
- Memory usage
- Requests per seconds
- Requests error rate 
- Latency of successful requests
- Latency of errored requests

Closes #525

### DIFF
--- a/infrastructure/modules/cluster/dashboard.tf
+++ b/infrastructure/modules/cluster/dashboard.tf
@@ -28,9 +28,9 @@ resource "azurerm_dashboard" "doc-index-updater-dashboard" {
                 "name": "ComponentId",
                 "value": {
                   "SubscriptionId": "${data.azurerm_subscription.current.subscription_id}",
-                  "ResourceGroup": "mhra-products-development",
-                  "Name": "mhra-products-cluster-analytics-gv55",
-                  "ResourceId": "/subscriptions/${data.azurerm_subscription.current.subscription_id}/resourcegroups/mhra-products-development/providers/microsoft.operationalinsights/workspaces/mhra-products-cluster-analytics-gv55"
+                  "ResourceGroup": "${var.resource_group_name}",
+                  "Name": "${azurerm_log_analytics_workspace.cluster.name}",
+                  "ResourceId": "/subscriptions/${data.azurerm_subscription.current.subscription_id}/resourcegroups/${var.resource_group_name}/providers/microsoft.operationalinsights/workspaces/${azurerm_log_analytics_workspace.cluster.name}"
                 }
               },
               {
@@ -77,7 +77,7 @@ resource "azurerm_dashboard" "doc-index-updater-dashboard" {
               },
               {
                 "name": "PartSubTitle",
-                "value": "mhra-products-cluster-analytics-gv55"
+                "value": "${azurerm_log_analytics_workspace.cluster.name}"
               },
               {
                 "name": "resourceTypeMode",
@@ -100,7 +100,7 @@ resource "azurerm_dashboard" "doc-index-updater-dashboard" {
             "settings": {
               "content": {
                 "PartTitle": "Request error rate",
-                "PartSubTitle": "mhra-products-cluster-analytics-gv55"
+                "PartSubTitle": "${azurerm_log_analytics_workspace.cluster.name}"
               }
             },
             "asset": {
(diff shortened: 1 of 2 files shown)


infrastructure/modules/products-web/main.tf
```diff
@@ -1,10 +1,4 @@
 
-resource "azurerm_storage_container" "products_website" {
```
rjkerrison, line 2: are we sure we want to remove this `products_website`?

infrastructure/modules/cluster/dashboard.tf
```diff
@@ -0,0 +1,244 @@
...
+            "y": 0,
+            "colSpan": 6,
+            "rowSpan": 4
+          },
+          "metadata": {
+            "inputs": [
+              {
+                "name": "ComponentId",
+                "value": {
+                  "SubscriptionId": "bec11470-1346-4cdd-af2e-ce1f360671a1",
```
rjkerrison, line 28: is this a secret?

infrastructure/modules/cluster/dashboard.tf
```diff
@@ -0,0 +1,244 @@
+resource "azurerm_dashboard" "doc-index-updater-dashbaord" {
```
rjkerrison, line 1: name typo `dashbaord` should be `dashboard`

infrastructure/modules/products-web/main.tf
sgrowe: Yeah, ever since we [added the static website block](https://github.com/MHRA/products/pull/571/files#diff-02147329094c80b8475c3f2e8ff56844) it looks like Azure creates the `$web` container on it's own. Will add it back if that turns out not to be the case

infrastructure/modules/cluster/dashboard.tf
sgrowe: You can't actually do anything with a subscription ID on it's own so I would say no, but I would agree that it's preferable to keep it private. This is should all be parameterised now so prod subscription ID will never be made public.

[More discussion](https://stackoverflow.com/a/45663968)

Issue #601, opened 2020-04-01
Delete service treats every error like a job error


rjkerrison:
**Describe the bug**

The doc-index-updater create service, sensibly, retries when it encounters unknown errors.

Delete throws a hissy fit, and cancels the job forever.

**To Reproduce**

Set any key or name used by the Delete service to an incorrect value, or go offline for a minute or two. 

**Expected behavior**
Delete jobs are retried until the dead letter count is hit.

**Actual behaviour**
Watch as every delete job immediately becomes an error.

**Screenshots**

**Additional context**

# QA information
unit tests : make test

**For doc not found in index** 

1. scripts/delete.sh 123123123123
2. (assumes that id doesn't exist)
3. wait for the lease lock to expire
4. see that it is not retried

**For other errors**

1. find a way to create another error - maybe wrong search index api key?
2. scripts/delete.sh REAL_ID_HERE
3. wait for the lease lock to expire
4. see that it is retried

majikandy:
@rjkerrison Do you have an example of something you think should retry? because something like 404 shouldn't retry, right ?

majikandy:
**Things that should be retried**
404 - Since we get this currently from permissions error
401 - Fix api keys in env and this message is recoverable
503 - service unavailable, can recover when it comes back

**Things that should not be retried**
Document not found in index - should this even be an error scenario?

craiga:
I can't think of anything else which shouldn't be retried, but can think of these error conditions:

* Blob described in search index not in storage (this might be 404?)
* We don't have permission to delete a blob (this might be 401?)

FYI, there's also https://github.com/MHRA/products/issues/561 which is tangentially related. It might be worth taking a look at that while you're looking at this?

sgrowe:
Whilst trying to QA this I hit upon a bug where if I make an API request to delete a document and it fails to add the message to the service bus queue it still returns a 200 success response code.

Example curl request:
```
curl -vvv -X DELETE --user username:password http://localhost:8000/documents/CON1581657002878
*   Trying ::1...
* TCP_NODELAY set
* Connection failed
* connect to ::1 port 8000 failed: Connection refused
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 8000 (#0)
* Server auth using Basic with user 'username'
> DELETE /documents/CON1581657002878 HTTP/1.1
> Host: localhost:8000
[...]

PR #632, opened 2020-04-07
Delete remove from service bus for not found ONLY

majikandy:
# A nice friendly title for what this PR does

Fixes issue #601 

### Acceptance Criteria

- [ ] Remove from service bus for not found in index
- [ ] keep it there for retries for any other error

### Testing information
unit tests : `make test`

**For doc not found in index **
1. `scripts/delete.sh 123123123123`
(assumes that id doesn't exist)
2. wait for the lease lock to expire
3. see that it is not retried

**For other**
1. find a way to create another error - maybe wrong search index api key?
2. `scripts/delete.sh <a real id here>`
3. wait for the lease lock to expire
4. see that it is retried

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -181,15 +191,31 @@ mod test {
...
+    #[test]
+    fn recoverable_error_during_delete_does_not_remove_message_from_servicebus() {
+        let state_manager = given_a_state_manager();
+        let mut removeable_message = given_we_have_a_delete_message();
+        let error = given_an_error_has_occurred(anyhow!("Any other error"));
 
-        let result = when_we_handle_the_error(&mut removable_message, error, state_manager);
+        let result = when_we_handle_the_error(&mut removeable_message, error, state_manager);
 
         assert!(result.is_ok());
```
majikandy, line 215: I feel that wrapping these 2 lines in a then_xxxx() would be good as then the full GWT is clear - but it isn't ridicuous for the assert to be more forefront like this

medicines/doc-index-updater/src/delete_manager/errors.rs
```diff
@@ -0,0 +1,23 @@
...
+    content_id: String,
+}
+
+impl DocumentNotFoundInIndex {
+    pub fn for_content_id(content_id: String) -> DocumentNotFoundInIndex {
+        DocumentNotFoundInIndex { content_id }
+    }
+}
+
+impl fmt::Display for DocumentNotFoundInIndex {
```
rjkerrison, line 13: there's a shorthand for this by deriving from `thiserror::Error`
https://github.com/MHRA/products/blob/2b305e1f628b862aa63f1cee77e158c1ab657b98/medicines/doc-index-updater/src/create_manager/sftp_client.rs#L5-L9

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -181,15 +191,31 @@ mod test {
     }
 
     #[test]
-    fn test_an_error_removes_delete_message() {
-        let state_manager = TestJobStatusClient {};
+    fn not_found_error_during_delete_removes_message_since_no_need_to_retry() {
+        let state_manager = given_a_state_manager();
+        let mut removeable_message = given_we_have_a_delete_message();
+        let error = given_an_error_has_occurred(anyhow!(
```
rjkerrison, line 197: I'd create a method called `given_the_document_was_not_found` which returns the error we'd expect in that scenario

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -181,15 +191,31 @@ mod test {
...
+
+        assert!(result.is_ok());
+        assert_eq!(removeable_message.is_removed, true, "Didn't remove message");
+    }
+
+    #[test]
+    fn recoverable_error_during_delete_does_not_remove_message_from_servicebus() {
+        let state_manager = given_a_state_manager();
+        let mut removeable_message = given_we_have_a_delete_message();
+        let error = given_an_error_has_occurred(anyhow!("Any other error"));
```
rjkerrison, line 211: can this be `given_an_unknown_error_has_occurred` which returns an `anyhow` error we wouldn't expect to be handled

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -112,8 +117,7 @@ pub async fn get_blob_name_from_content_id(
             return Ok(result.metadata_storage_name);
         }
     }
-    Err(anyhow!(format!(
-        "Cannot find document with content ID {}",
+    Err(anyhow!(errors::DocumentNotFoundInIndex::for_content_id(
```
rjkerrison, line 120: I think we can and should unit test this. It'll require an adapter for the `AzureSearchClient`, but it will be worth it to test that the right error is representing the _blob not found_ scenario.

medicines/doc-index-updater/src/delete_manager/errors.rs
rjkerrison: let's leave this out for now

medicines/doc-index-updater/src/delete_manager/mod.rs
rjkerrison: we can defer this too

majikandy:
This is functional in the behaviour we need, so merging now, but another PR will be coming soon with testing around the raising of the error that this PR handles in order to remove the message from servicebus(or not)

Issue #621, opened 2020-04-06
Search results page uses API


KaylieGreen1607:
### User want
As a 3rd party user I would like to able to access search information through an API, so that I don't have to scrape the website and have easy access to the information.

### Acceptance Criteria

**Customer acceptance criteria** 

- [x] A third party user can access the documents API. It should be self-documenting to a user who is familiar with GraphQL.
- [x] Search results page of the site can load a complete list of documents for a given search term, as is the current behaviour.
- [x] Document lists can be filtered by document type, as is the current behaviour.

**Technical acceptance criteria**

- [x] The public-facing web site makes use of the documents API to retrieve a list of documents which are relevant to a search term.
- [x] The public-facing web site makes use of the documents API to filter those lists of documents.
- [x] The function of the web site should be otherwise unchanged.
- [x] The API and doc index updater should share an implementation of the rust Azure search client.
- [x] The Cypress tests ~are modified~ are added to reflect the use of the API. (Old ones can stay with the feature flag off until we remove the old direct-to-Azure features.)

**Size**
M

**Value**

**Effort**

### Exit Criteria met 
- [x] Backlog
- [x] Discovery
- [x] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

majikandy:
## QA
Results fail to load when using GraphQl

@majikandy   9:24 AM
https://mhraproductsnonprod.z33.web.core.windows.net/search/?search=caffeine&page=1&useGraphQl=true

400 bad request for the graphql call

@rjkerrison  9:25 AM
`{"errors":[{"message":"Variable \"documentTypes\" of type \"[String!]\" used in position expecting type \"[DocumentType!]\"","locations":[{"line":2,"column":57},{"line":3,"column":79}]}]}`

Suggest when we fix we add some kind of protection for the contract between web and api.

rjkerrison:
Thanks @majikandy. That bug should be fixed by #753.

majikandy:
## QA 2

```
 ➜  ~ curl https://mhraproductsnonprod.z33.web.core.windows.net/search/\?search\=caffeine\&page\=1\&useGraphQl\=true > graphql.txt
 ➜  ~ curl https://mhraproductsnonprod.z33.web.core.windows.net/search/\?search\=caffeine\&page\=1 > search.txt  
 ➜  ~ diff graphql.txt search.txt 
 ➜  ~ 
```
PASS

majikandy:
[...]

Issue #620, opened 2020-04-06
Substance page uses API


KaylieGreen1607:
### User want

As a 3rd party user I would like to able to access substance information through an API, so that I don't have to scrape the website and have easy access to the information.

### Acceptance Criteria

**Customer acceptance criteria** 

- [ ] A third party user can access the products API. It should be self-documenting to a user who is familiar with GraphQL.
- [ ] Substance page of the site can load a complete, alphabetised list of products for a given substance, as is the current behaviour

**Technical acceptance criteria**

- [ ] The public-facing web site makes use of the products API to retrieve a list of products which use a substance.
- [ ] The function of the web site should be otherwise unchanged.
- [ ] The API and doc index updater should share an implementation of the rust Azure search client.
- [ ] The Cypress tests are modified to reflect the use of the API.
- [ ] ~~Self-documenting API is linked to from GitHub so it can be discovered by a third-party developer.~~ (see #701)

**Data acceptance criteria**

- [ ] ~~The fact the API exists should be publicised somewhere, so people know where to access it.~~ (see #701)

**Testing acceptance criteria** 
- [ ] https://mhraproductsnonprod.z33.web.core.windows.net/substance/?substance=CODEINE%20PHOSPHATE%20HEMIHYDRATE&useGraphQl=true fetches products for that substance from graphQl

**Size**
M

**Value**

**Effort**

### Exit Criteria met 
- [x] Backlog
- [x] Discovery
- [x] DUXD
- [x] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

craiga:
Now that the search client has been split into a separate crate, there are now two parallelizable pieces of work:

 * making use of the new search client crate in the API; and
 * making use of the API from the frontend.

majikandy:
QA

self documenting part was good and led to easily using graphiql and then supplying this query

```
{
  substance(name: "CAFFEINE") {
    products
  }
}

```

it then fails to actually run the query and get results -> this is the error

```
"upstream connect error or disconnect/reset before headers. reset reason: connection termination"
```

devtools
```
Summary
URL: https://medicines-api.test.mhra.gov.uk/graphql
Status: 503
Source: Network
Address: 51.132.171.245:443
Initiator: 
graphiql:1:882


Request
:method: POST
:scheme: https
:authority: medicines-api.test.mhra.gov.uk
:path: /graphql
Accept: application/json
Content-Type: application/json
Origin: https://medicines-api.test.mhra.gov.uk
[...]

Issue #403, opened 2020-02-17
Drug Index Page uses API


ainemckay:
### User want
As a 3rd party user I would like to able to access drug index information through an API, so that I don't have to scrape the website and have easy access to the information.

Customer acceptance criteria

 A third party user can access the documents API. It should be self-documenting to a user who is familiar with GraphQL.

 Product page of the site can load a complete list of documents for a given product, as is the current behaviour.

Technical acceptance criteria

The public-facing web site can make use of the documents API to retrieve a list of documents which are relevant to a product. This is enabled with a query string flag and a hosts file entry.

The public-facing web site makes use of the substances API to filter those lists.
The function of the web site should be otherwise unchanged.

 The API and doc index updater should share an implementation of the rust Azure search client.

 The Cypress tests are modified to reflect the use of the API.

**Size**
M

**Value**

**Effort**

### Exit Criteria met 
- [ ] Backlog
- [ ] Discovery
- [ ] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

majikandy:
# QA - on non prod
When using  &useGraphQl=true it isn't returning the same results as without

To reproduce...  
Click on Z
see that 4 results are there with 12 docs in total
```
Z

ZIDOVUDINE (2 files)
ZOLEDRONIC ACID (2 files)
ZOLMITRIPTAN (2 files)
ZONISAMIDE (6 files)
```
now add &useGraphQl=true to the url
see that 3 results are there with only 8 docs in total
```
Z

ZOLEDRONIC ACID (2 files)
ZOLMITRIPTAN (2 files)
ZONISAMIDE (6 files)
```
cc @sgrowe

sgrowe:
@majikandy 

Good find 👍  

Fixed that in #729 and ready for QA again

majikandy:
QA 
This is now behaving correctly.

There is still an issue where documents aren't found on click through where the drug index page shows products with commas as only the text up to the comma. Described in #738 

It is ok for this card to go forward, but the caveat is that we shouldn't deploy to prod unless either #738 is fixed for non GraphQL query or GraphQL is made the default api to use.

Issue #401, opened 2020-02-17
Product Page uses API


ainemckay:
### User want

**Customer acceptance criteria** 

- [x] API is in a state where it could be used easily by a third party (e.g. NHS developer)

**Technical acceptance criteria**

- [x] API allows for queries of documents for a specific product
- [ ] A list of documents should be returned in the same order as the current website
- [x] Product name should exactly match the product associated with each document.
- [ ] Documents can be filtered by one or more document type (i.e. PIL, PAR, SPC)
- [ ] The output of the API should include pagination.
- [ ] website uses this when feature query param `useGraphQl=true` set 

This relates to #454.

**Data acceptance criteria**

**Testing acceptance criteria** 

- [ ] https://mhraproductsnonprod.z33.web.core.windows.net/product/?product=CO-CODAMOL%208MG%2F500MG%20TABLETS&useGraphQl=true fetches results from graphQl

**Size**
M

**Value**

**Effort**

### Exit Criteria met 
- [ ] Backlog
- [ ] Discovery
- [ ] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

rjkerrison:
Currently failing due to an issue where a variable named `offset` is being called `skip` by the website. Will be fixed by #754 .

majikandy:
# QA
Working as expected with and without useGraphQl=true

GraphQl returns the data in a slightly different order but the website returns it in the same order with and without the flag and the client side filtering by doc type still works.

PR #641, opened 2020-04-08
Extract Search Client as a separate crate.

m-doughty:
# Extract Search Client as a separate crate

_Make the Search Client a reusable component which can be shared between Rust code bases which want to search the index._

relates to #401, #403, #620, #621.

### Acceptance Criteria

- [x] The Azure Search Client crate should be self-contained;
- [x] Azure Search Client code should be removed from the doc-index-updater sub-repo;
- [x] doc-index-updater should work as expected using the Azure Search Crate.

### Testing information

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


medicines/search-client/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "search-client"
+version = "0.1.0"
+authors = ["Matt Doughty <matt.doughty@mhra.gov.uk>", "Craig Anderson <craig.anderson@red-badger.com>"]
```
majikandy, line 4: is this intentional?

m-doughty: Is what intentional?

majikandy: having specific developer names

m-doughty: I thought we had those in all of our Cargo.toml files?

craiga: You need an email address in your `Cargo.toml` in order to publish to crates.io.

.github/workflows/doc-index-updater-branch.yaml
```diff
@@ -21,6 +21,10 @@ jobs:
       - name: Clone Repo
         uses: actions/checkout@v2
 
+      - name: Test for RC release
+        working-directory: ./medicines/doc-index-updater
+        run: cmd=`cat Cargo.toml | grep mhra_products_search_client | sed -e 's/"//' | sed -e 's/"//'` && if [[ $cmd == *"-rc"* ]];then echo rc dependencies not allowed for this package; exit 1; fi
```
majikandy, line 26: can be simplified to:-
`case $(grep <Cargo.toml mhra_products_search_client) in *"-rc"*) echo "Only stable versions of mhra_products_search_client should be used"; exit 1 ;; esac`

StuartHarris:
Please see my other [comment](https://github.com/MHRA/products/pull/650#issuecomment-612076592)

craiga:
> I'm not sure that we should be publishing like this to crates.io. As they stand, these crates don't seem to be generally useful right now (especially as they have mhra in the title) and so IMO don't belong there (yet).

@StuartHarris Here's a brief history of how we got here:

* @m-doughty, @majikandy and I moved the search client into its own crate in `medicines/search-client` so it could be shared between doc index updater and the API.
* We found that didn't work inside Docker containers, as Docker can only use files which are inside the directory the Dockerfile lives in.
* We talked about moving Dockerfiles about, changing build scripts, etc., but thought that pushing the shared package was a more elegant solution than polluting the monorepo with Dockerfiles a level about where they should be.

I agree that this isn't a great solution, so I'd be interested in hearing your thoughts on this. Do you think moving Dockerfiles would be a better solution than pushing this crate? Do you know of any other solutions to this problem?

I also agree that we shouldn't be pushing additional temp packages. I dropped the ball by not giving @majikandy permission to push to that crate before leaving for the long weekend.

StuartHarris:
OK, thanks. We should definitely find another solution than pushing to crates.io. Those 3 crates will now be there forever, and that's not cool. We should only use crates.io for crates that we want to share with the world.
I would suggest either moving the `Dockerfile` up, or, even better, using `tar` to send the relevant context to the `docker build`. e.g. `tar cvf ../whatever | docker build -`

medicines/doc-index-updater/Cargo.toml
```diff
@@ -20,6 +20,7 @@ md5 = "0.7.0"
 redis = { version = "0.15.1", features = ["tokio-rt-core"] }
 regex = "1.3.1"
 reqwest = { version = "0.10.4", features = ["json"] }
+search_client =  { git = "https://github.com/MHRA/products", branch = "createable-deleteable-into-search-crate" }
```
rjkerrison, line 23: I think this is the problem that [Cargo workspaces][1] solve. It seems weird for a repo to reference itself like this.

[1]: https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html

majikandy: Ok cool, we'll try it and compare

majikandy: the issue with workspaces is that you then reference the search-client package by doing `search_client = { path = ../search-client }` and then `make docker-build` won't work as the search-client is below the root of where the Dockerfile is.   So we could change all that too and all the build/release process but potentially a lot of changes (maybe too much right now) to achieve this?

majikandy: changed to use tar to pipe both the `doc-index-updater` and the `search-client` into `docker build` (via `make docker-build`)

majikandy:
Now implemented using tar to pass doc-index-updater with dependency as a single tarball piped into docker build. No longer needs crates.io or GitHub url reference. Just a simple .. folder path

.github/workflows/search-client-master.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+    steps:
+      - name: Clone Repo
+        uses: actions/checkout@v2
+        with:
+          path: products
+
+      - name: Test
+        working-directory: ./medicines/search-client
+        run: cargo test
```
sgrowe, line 24: `search-client-branch.yaml` and `search-client-master.yaml` do the same thing so we could potentially just have one of them

medicines/doc-index-updater/Makefile
```diff
@@ -32,13 +32,12 @@ ci-branch: docker-build ## Build branch in CI
 ci-master: ci-branch docker-push ## Build master in CI
 
 .PHONY: docker-build
-docker-build: ## Build and tag Docker image
-	docker build \
+docker-build: ## Build and tag Docker image (with search-client create as dependency)
+	tar cv --exclude=target Dockerfile -C .. ./doc-index-updater ./search-client | docker build - \
```
sgrowe, line 36: Would it perhaps be simpler just to move the docker file up one folder (and maybe rename it to something like `Dockerfile.doc-index-updater`) so then you could just do `docker build -f Dockerfile.doc-index-updater`?

medicines/search-client/docs/development-and-releasing.md
```diff
@@ -0,0 +1,15 @@
+# Developing and Releasing this Package
+
+## Release Candidates
+
+To publish a new, release candidate version of this package:
+
+1. Set a new version number in `Cargo.toml` with an `-rc` suffix (for example, `1.2.2` is incremented to `1.2.3-rc1`, `1.2.3-rc4` is incremented to `1.2.3-rc5`).
```
sgrowe, line 7: Is this file still needed if we're not pushing to crates.io anymore?

majikandy: Nope, good point.

medicines/doc-index-updater/Makefile
majikandy: We did consider that but went with this so that the Dockerfile for doc-index-updater is actually with it - could get messy in the folder above if more appear over time.

.github/workflows/search-client-master.yaml
majikandy: True, just matches the pattern of the other packages and allows it to diverge.

PR #646, opened 2020-04-09
Createable deleteable

majikandy:
# Testability extraction
pulls out behaviours of AzureSearchClient into traits

### DIFF
--- a/medicines/doc-index-updater/src/search_client/mod.rs
+++ b/medicines/doc-index-updater/src/search_client/mod.rs
@@ -106,8 +106,18 @@ impl Searchable for AzureSearchClient {
     }
 }
 
-impl AzureSearchClient {
-    pub async fn delete(
+#[async_trait]
+pub trait Deletable {
+    async fn delete(
+        &self,
+        key_name: &str,
+        value: &str,
+    ) -> Result<AzureIndexChangedResults, anyhow::Error>;
+}
+
+#[async_trait]
+impl Deletable for AzureSearchClient {
+    async fn delete(
         &self,
         key_name: &str,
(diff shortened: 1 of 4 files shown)


medicines/doc-index-updater/src/search_client/mod.rs
```diff
@@ -106,8 +106,18 @@ impl Searchable for AzureSearchClient {
...
+pub trait Deletable {
+    async fn delete(
+        &self,
+        key_name: &str,
+        value: &str,
+    ) -> Result<AzureIndexChangedResults, anyhow::Error>;
+}
+
+#[async_trait]
+impl Deletable for AzureSearchClient {
```
rjkerrison, line 119: is there another name we could use? this reads to me like the `AzureSearchClient` will be deleted
also seems like ` Searchable + Deletable + Createable` can be combined, so perhaps `IndexSearcher` or `IndexClient` or something is more appropriate

majikandy: went for Search, DeleteIndexEntry, CreateIndexEntry as it read quite nicely from the client code

This is updated in PR #641  that pulls out the search client so I will abandon this PR

majikandy:
These changes are now merged in PR #641 so abandoning this as it will get merged to master via that PR

medicines/doc-index-updater/src/search_client/mod.rs
majikandy: And yes they could be combined, and will need to be when we change process_message as that dependency can be injected instead of ::factory() call at the start of the function -

PR #649, opened 2020-04-09
use async sftp client

StuartHarris:
Use an async SFTP client instead of blocking the current thread.


TimboTambo:
Confirmed this works locally

PR #650, opened 2020-04-09
Createable deleteable into search crate

majikandy:
Same as PR #646 except merging to the search client crate extraction branch (if we choose to go with that approach)

I didn't have permission on crates.io ->`mhra_products_search_client` so had to temporarily push to the same name+  `_temp` in order to reference it in the client code.

Note: I think this would be confusing a tricky for a new developer to do.

### DIFF
--- a/medicines/doc-index-updater/Cargo.toml
+++ b/medicines/doc-index-updater/Cargo.toml
@@ -20,7 +20,7 @@ md5 = "0.7.0"
 redis = { version = "0.15.1", features = ["tokio-rt-core"] }
 regex = "1.3.1"
 reqwest = { version = "0.10.4", features = ["json"] }
-mhra_products_search_client_temp = "0.0.2-rc2"
+search_client =  { git = "https://github.com/MHRA/products", branch = "createable-deleteable-into-search-crate" }
 serde = "1.0.105"
 serde_derive = "1.0.105"
 serde_json = "1.0"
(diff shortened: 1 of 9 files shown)


StuartHarris:
I'm not sure that we should be publishing like this to [crates.io](https://crates.io). These crates are not generally useful (especially as they have `mhra` in the title) and so IMO don't belong there. You can't delete crates from crates.io and so these will remain there forever (the best you can do is yank them, but they will remain). Additionally, I don't think that not having permissions to push to an existing crate should result in additional 'temp' crates being published which effectively just pollute a valuable resource.

StuartHarris:
Sorry, closed by accident :-)

majikandy:
Changed to not use crates.io for search-client

medicines/doc-index-updater/Cargo.toml
```diff
@@ -20,7 +20,7 @@ md5 = "0.7.0"
 redis = { version = "0.15.1", features = ["tokio-rt-core"] }
 regex = "1.3.1"
 reqwest = { version = "0.10.4", features = ["json"] }
-mhra_products_search_client = "0.0.1"
+search_client =  { git = "https://github.com/MHRA/products", branch = "createable-deleteable-into-search-crate" }
```
rjkerrison, line 23: I think this is the problem that [Cargo workspaces][1] solve. It seems weird for a repo to reference itself like this.

[1]: https://doc.rust-lang.org/book/ch14-03-cargo-workspaces.html

PR #651, opened 2020-04-11
Fehler

StuartHarris:
Converts `doc-index-updater` to use [`fehler`](https://github.com/withoutboats/fehler) in order to understand whether this simplifies error handling. It introduces the `throws` attribute to avoid having to return a `Result` everywhere. This can make functions easier to write and to read.

We should agree as a team whether to merge this or not as it may not be to everyone's taste :-)

Two caveats:
- I couldn't get it to work inside `async_trait` implementations (there are quite a few)
- I didn't convert functions that contain infinite loops (there are 2), as we get an `unreachable code` warning


rjkerrison:
Might be worth a call about this. My initial thought is that it's not very `rust`-y, but maybe it's just ahead of the curve. I do find it useful to see which methods are infallible, and this would obfuscate that slightly.

m-doughty:
:-1:

Definitely worth doing the investigation, but I think this isn't particularly rust-y as @rjkerrison said, and I think the added difficulty reading and understanding function signatures (especially given it isn't everywhere) takes more away from development efficiency than the less verbose error handling adds.

StuartHarris:
Closing this as it seems there were more against than for (and I'm still not even sure myself) :-)

Issue #658, opened 2020-04-15
Job set to error status even though it was being retried 


TimboTambo:
**Describe the bug**
A delete job failed because it was unable to reach the search service, the job status was changed to `Error(: error sending request for url (https://mhraproductsnonprod.search.windows.net/indexes/products-index/docs?api-version=2019-05-06&highlight=content&queryType=full&%40count=true&%40top=10&%40skip=0&search=CON1563509585664&scoringProfile=preferKeywords): error trying to connect: unexpected EOF)`. 
It was then retried and was successful.
The consumer (Sentinel batch process) stopped checking for updates after the first error status.

**To Reproduce**
Check logs with following query:
```let correlationId = "f9ec57a6-3d54-43d6-a1ee-924117f2b72b";
let timeframeFrom = totimespan(7d);
let timeframeTo = totimespan(5d);
let clusterId = '/subscriptions/bec11470-1346-4cdd-af2e-ce1f360671a1/resourceGroups/adazr-rg-1001/providers/Microsoft.ContainerService/managedClusters/non-prod';
let ContainerIdList = KubePodInventory
| where TimeGenerated > now() - timeframeFrom and TimeGenerated < now() - timeframeTo
| where ContainerName contains 'doc-index-updater'
| where ClusterId =~ clusterId
| distinct ContainerID;
ContainerLog
| where TimeGenerated > now() - timeframeFrom and TimeGenerated < now() - timeframeTo
| where ContainerID in (ContainerIdList)
| project LogEntrySource, LogEntry, TimeGenerated, Computer, Image, Name, ContainerID
| order by TimeGenerated desc
| render table
| extend message_ = tostring(parse_json(tostring(parse_json(LogEntry).fields)).message) 
| where parse_json(tostring(parse_json(LogEntry).span)).correlation_id == correlationId```

**Expected behavior**
The job status should not be updated to `error` if the message is not removed from the queue.

**Screenshots**
N/A

**Additional context**
N/A

Issue #653, opened 2020-04-14
Delete can remove document from search index but not storage, making it impossible for Sentinel to delete it


craiga:
From @TimboTambo:

> There were around 30 errors from the batch run test last week that occurred with deleting files. Seems like in all cases the doc to delete was found in the search index, removed from the search index, failed when removing from the storage container, was retried and then the doc could no longer be found in the search index so was set to Error(: Cannot find document with ID …)

The following Azure log query brings up an example of this happening:

```
let correlationId = "bd582bbf-6a76-4708-9440-10aefc726d89";
let timeframeFrom = totimespan(6d);
let timeframeTo = totimespan(4d);
let clusterId = '/subscriptions/bec11470-1346-4cdd-af2e-ce1f360671a1/resourceGroups/adazr-rg-1001/providers/Microsoft.ContainerService/managedClusters/non-prod';
let ContainerIdList = KubePodInventory
| where TimeGenerated > now() - timeframeFrom and TimeGenerated < now() - timeframeTo
| where ContainerName contains 'doc-index-updater'
| where ClusterId =~ clusterId
| distinct ContainerID;
ContainerLog
| where TimeGenerated > now() - timeframeFrom and TimeGenerated < now() - timeframeTo
| where ContainerID in (ContainerIdList)
| project LogEntrySource, LogEntry, TimeGenerated, Computer, Image, Name, ContainerID
| order by TimeGenerated desc
| render table
| extend message_ = tostring(parse_json(tostring(parse_json(LogEntry).fields)).message) 
| where parse_json(tostring(parse_json(LogEntry).span)).correlation_id == correlationId
```

craiga:
Currently the order of operations in `process_message` is:

* get blob name from search index
* delete from search index
* delete from blob

We should delete from blob before deleting from search index.

PR #659, opened 2020-04-15
Delete blob before deleting entry from search index. Addresses #653

craiga:
Delete blob before deleting entry from search index.

Addresses #653.

### Acceptance Criteria

- [ ] If deleting from blob storage fails, the delete job should be able to be retried.

### Testing information

To test, I:

* put a document into the system;
* changed the environment variable `STORAGE_CONTAINER` to a garbage value
* restarted the doc index updater
* sent a request to delete the document
* saw that the delete failed
* saw that the delete was retried
* changed `STORAGE_CONTAINER` back to its original value
* restarted the doc index updater
* saw that the delete was retried and succeeded

Note that the job status is updated after every attempt to process the delete message, this should be fixed in #658


craiga:
@majikandy How could this order be checked? I get that the individual methods called from `process_message` should be unit tested, but it seems like overkill to unit test `process_message` itself, especially since it's the method responsible for creating the search and storage clients in the first place.

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -108,6 +107,9 @@ pub async fn process_message(message: DeleteMessage) -> Result<Uuid, ProcessMess
         &storage_container_name
```
TimboTambo, line 107: I believe this will error if the blob has already been deleted? So there will still be an error if the first attempt deletes the blob but fails to delete from the index and then it's retried.

craiga: Good point. I'll look into it.

m-doughty: Best way to achieve consistent response is probably to treat deleting blob & deleting index as separate operations:

- Receive message to delete document
- Look up in index
- Delete blob
- If successful, dispatch message to delete in index
- Receive message to delete index
- Delete index
- Mark as done

Thoughts?

craiga: @TimboTambo Just to confirm, is the desired behaviour here:

 * if blob exists, delete it and remove it from index
 * if blob does not exist, emit a warning and remove if from index
 * if an error* occurs, raise an error and retry

?

<small>* — an error which isn't "blob does not exist"; for example, a the blob has an open lease, a permission error with the index, etc.</small>

TimboTambo: Agree that a more robust way of making idempotent would be to split out into two separate tasks with another message. As a short-term fix, I think @craiga's mentioned flow would be sufficient to prevent the observed issue from re-occurring.

m-doughty: I'm not sure what log behavior is expected, but Accenture are expecting an **error** response if they get the job and the blob doesn't exist. I'm not sure that we need to give them the same if the index couldn't be deleted.

m-doughty:
> This can be bought under test and the order checked - the traits introduced can help with this but we'd need the search-client crate extraction PR merged first

I agree, but I don't think it's in the scope of this ticket (and these tickets are time-sensitive so I don't think we should be increasing scope hugely unless it's necessary to finish the task).

medicines/doc-index-updater/src/create_manager/models.rs
```diff
@@ -366,7 +369,10 @@ mod test {
                 doc_type: DocumentType::Spc,
                 title: SanitisedString::from("Some SPC".to_string()),
                 pl_number: "[\"PL123450010\"]".to_string(),
-                product_names: VecSanitisedString::from(vec!["GENERIC PARACETAMOL".to_string(), "SPECIAL PARACETAMOL".to_string()]),
+                product_names: VecSanitisedString::from(vec![
+                    "GENERIC PARACETAMOL".to_string(),
+                    "SPECIAL PARACETAMOL".to_string()
+                ]),
```
craiga, line 375: These changes due to `cargo fmt`.

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -94,19 +95,37 @@ pub async fn process_message(message: DeleteMessage) -> Result<Uuid, ProcessMess
...
+                    );
+                    Ok(())
+                } else {
+                    tracing::error!("Error deleting blob: {:?}", e);
+                    Err(anyhow!("Couldn't delete blob {}: {:?}", &blob_name, e))
+                }
+            }
+            e => Err(anyhow!("Couldn't delete blob {}: {:?}", &blob_name, e)),
+        },
+    }?;
```
craiga, line 125: All of this error handling should be moved into `delete_blob` and brought under test, but doing that would take me longer than we have left in today.

m-doughty: Does this correctly set the job status to an error if the blob couldn't be found?

craiga: No. It assumes that a blob not being found is OK.

I don't think the blob not being found in blob storage when it is found in the search index is an error.

TimboTambo: @m-doughty yes it looks like it would. This was the previous behaviour, when the error was mapped to the `anyhow!("Couldn't delete blob {}", &blob_name)` error - the status updated to this.

This looks ok as a stop-gap fix to me, with a clean-up to follow.

m-doughty: > No. It assumes that a blob not being found is OK.
> 
> I don't think the blob not being found in blob storage when it is found in the search index is an error.

This discussion was had with Accenture, they are requesting an error state when the document they attempted to delete wasn't found in Blob storage.

TimboTambo:
I think there are two ways of thinking about this:
1. If we find the doc in the index but not in the storage container then we assume that there was an original attempt at deleting this document but the delete from the index was not successful, so we don't error but retry deleting from the index. We only error if we don't find the document in the index to begin with.
2. We error if the blob isn't found but not if the entry cannot be deleted from the search index, so that we don't even reach the situation in option 1.
Since option 1 is already implemented and we'll be updating this bit of code anyway with a more idiomatic solution then I think we should accept the current change.

medicines/doc-index-updater/src/delete_manager/mod.rs
majikandy: Unit of Work pattern is the normal way of achieving atomicity across different services (index and blob) isn't it? treating as separate seems a mistake as yes you won't get errors, you'll just get inconsistencies.

majikandy:
@m-doughty when it is time sensitive, this is even more reason to prove the fix through unit tests.

@craiga perhaps process_message could have these constructs injected so that it isn't its responsibility to construct them and can therefore focus on just processing the message. 
Perhaps testing the order might be slightly tricky but certainly we can test they are called at all.

TimboTambo:
Replaced by #672

PR #660, opened 2020-04-15
Propagate the real redis error to the client instead of hiding it?

majikandy:
_Errors were hidden behind "Other(Server Error)"._

Raising this for discussion since I often experience forgetting to start Redis server and see Other (Server Error) and it feels like more information to the client call is better. (As long as no credentials or attack vectors exposed)


craiga:
Would be good to see some before and after comparisons as part of this PR.

majikandy:
Before
`Other (Server Error)`
After
`RedisIoError(Connection refused (os error 61))`

majikandy:
@m-doughty decent idea

majikandy:
added the var via make set-env too

PR #661, opened 2020-04-15
add hello world example

StuartHarris:
Adds a `hello-world` simple API as an example. 

There is currently just one route (https://hello-world-dev.test.mhra.gov.uk/hello/SomeName) which returns the text `Hello SomeName`.


hello-world/src/lib.rs
```diff
@@ -0,0 +1,26 @@
+use tide::{Request, Server};
+
+pub fn create_app() -> Server<()> {
+    let mut app = tide::new();
+    app.at("/hello/:name").get(hello_handler);
+    app
+}
+
+async fn hello_handler(req: Request<()>) -> tide::Result<String> {
+    let name = req.param("name")?;
```
sgrowe, line 10: Tide seems pretty neat overall but the lack of type-safe routing is a shame 😢

hello-world/tests/hello_world.rs
```diff
@@ -0,0 +1,24 @@
...
+use std::time::Duration;
+
+#[test]
+fn hello_world() -> anyhow::Result<()> {
+    task::block_on(async {
+        let server =
+            task::spawn(async { hello_world::create_app().listen("localhost:8080").await });
+
+        let client = task::spawn(async {
+            task::sleep(Duration::from_millis(100)).await;
```
sgrowe, line 12: Seems like a bit of a flakey test? But I guess the app is just for demoing deployment so no big deal

hello-world/src/lib.rs
StuartHarris: `req.param()` is generic over the type though, so I suppose the end result is still type safe (as in you could write `let uid : u32 = req.param("user-id")?;`)

hello-world/tests/hello_world.rs
StuartHarris: Yeah, I still haven't worked out the best way to do this. I was following what they did in their integration tests. Not ideal, and without the delay it does fail every now and then. I haven't seen it fail with the delay. I think what's worse is that parallel tests could clash on the port. You can use port `0` to get an ephemeral port, but no way to know what that port is :-(. These things will improve.

hello-world/src/lib.rs
sgrowe: Ah yeah I guess it is type-safe in that sense, I just meant though that if you typo `"name"` for example then it will still compile but  it looks like you will just get back a `None` at run-time (or an `Err`..?). Whereas in warp if you have a `String` parameter in the url you are guaranteed to always get a string back

StuartHarris: hmm, good point.

PR #662, opened 2020-04-15
State machine

### LINKED ISSUE #657: Error removing message from queue
**Describe the bug**
A delete job completes successfully but there is an error removing the message from the queue, so the job is performed again and fails because the file is no longer in the search index/storage container.

This happened for around 30/600 files so failure rate of ~5%.

The reported error is: `UnexpectedHTTPResult(UnexpectedHTTPResult { expected: [200], received: 404, body: "u003cErroru003eu003cCodeu003e404u003c/Codeu003eu003cDetailu003eThe lock supplied is invalid. Either the lock expired, or the message has already been removed from the queue. TrackingId:263452ee-dfff-4c56-a7f3-7f5ac1858d39_G0_B11, SystemTracker:doc-index-updater-non-prod:Queue:doc-index-updater-delete-queue, Timestamp:2020-04-08T16:41:31u003c/Detailu003eu003c/Erroru003e" }`

**To Reproduce**
Seems like a transient issue - would be difficult to reproduce. You can see the logs for several jobs that experienced this by running the following query with jobs:

Jobs
9c16de58-ea5f-4720-9234-7986ac7f057c
488be267-ad3e-4127-88f6-83248456b2ce
2d561a7c-fdf5-4837-bbc6-224cdceee676
dd2cbe06-7dfc-4eb7-ace4-7cdd6004e598
85fd35cd-2450-4b59-bc80-95c11c5f154f

Query
```let correlationId = "9c16de58-ea5f-4720-9234-7986ac7f057c";
let timeframeFrom = totimespan(7d);
let timeframeTo = totimespan(5d);
let clusterId = '/subscriptions/bec11470-1346-4cdd-af2e-ce1f360671a1/resourceGroups/adazr-rg-1001/providers/Microsoft.ContainerService/managedClusters/non-prod';
let ContainerIdList = KubePodInventory
| where TimeGenerated > now() - timeframeFrom and TimeGenerated < now() - timeframeTo
| where ContainerName contains 'doc-index-updater'
| where ClusterId =~ clusterId
| distinct ContainerID;
ContainerLog
| where TimeGenerated > now() - timeframeFrom and TimeGenerated < now() - timeframeTo
| where ContainerID in (ContainerIdList)
| project LogEntrySource, LogEntry, TimeGenerated, Computer, Image, Name, ContainerID
| order by TimeGenerated desc
| render table
| extend message_ = tostring(parse_json(tostring(parse_json(LogEntry).fields)).message) 
| where parse_json(tostring(parse_json(LogEntry).span)).correlation_id == correlationId```

**Expected behavior**
Messages are removed from the queue once completed so aren't retried.

**Screenshots**
N/A

**Additional context**
There is a second issue in that a successful job status is being overridden by an error status when the job fails the second time it is run. An easier fix may be to address that issue.

### PR DESCRIPTION
**Pull request by sgrowe**
Fixes #657 

Now we use a couple to lua scripts to make sure that:

- [ ] we only set a job's state to "Accepted" if it doesn't exist already
- [ ] we only change a job's state if it is currently "Accepted" or doesn't exist already
- [ ] (therefore) if a job's state is "Done" or "Error", we don't ever change its state

### CODE REVIEW

File: medicines/doc-index-updater/src/state_manager/redis.rs [hunk 1]
```diff
@@ -72,13 +72,16 @@ pub async fn get_from_redis(client: Client, id: Uuid) -> RedisResult<JobStatus>
...
     let mut con = client.get_async_connection().await?;
 
-    redis::cmd("SET")
-        .arg(id.to_string())
-        .arg(status.clone())
-        .query_async(&mut con)
-        .await?;
+    let lua_script = match status {
+        JobStatus::Accepted => include_str!("set_if_not_exists.lua"),
+        _ => include_str!("set_if_not_completed.lua"),
```
TimboTambo, line 77: This script will also not set if errored or in any other state so maybe `set_if_not_exists_or_accepted`?

PR #665, opened 2020-04-15
Small infra updates

sgrowe:
- Made `data.azurerm_subnet.cluster_nodes` explicitly depend on the cluster in terraform as it's `route_table_id` doesn't seem to be set until the cluster is up (I might be mistaken though?)
- Updates the script that sets the search `.env` file to also run the commands to set up the search datasource, index & indexer.


infrastructure/scripts/setup-search.sh
```diff
@@ -35,18 +35,32 @@ OUTPUT=$(terraform output -json)
 popd
 
 {
-    echo "API_ADMIN_KEY=$(echo "$OUTPUT" | jq .search_admin_key.value --raw-output)"
+    echo "API_ADMIN_KEY=\"$(echo "$OUTPUT" | jq .search_admin_key.value --raw-output)\""
```
TimboTambo, line 38: Might still be worth having the `update-search-env` as a separate script so that you can populate your local env even when the search service components are already configured?

PR #670, opened 2020-04-16
Readme wording change after following the steps

majikandy:


medicines/doc-index-updater/README.md
```diff
@@ -40,9 +40,9 @@ stunnel stunnel.conf
 make
 ```
 
-To get SFTP working locally, first set up a local user, with username and password, that you can use as a local SFTP server.
+To get SFTP working locally, first set up a local user, with a username that you can use as a local SFTP server (e.g. localsftpuser).
 
-Add this variables to an file `.env.overrides` using the env var `SENTINEL_SFTP_USERNAME`.
+Add `SENTINEL_SFTP_USERNAME=username` (where username is whatever you set in the step above, e.g. `SENTINEL_SFTP_USERNAME=localsftpuser`) to a file called `.env.overrides` which goes next to `.env` file (create the file if it doesn't exist).
```
rjkerrison, line 45: If you put these steps after the `make set-sftp-keys` step, `.env.overrides` will exist at this step.

medicines/doc-index-updater/README.md
```diff
@@ -40,9 +40,9 @@ stunnel stunnel.conf
 make
 ```
 
-To get SFTP working locally, first set up a local user, with username and password, that you can use as a local SFTP server.
+To get SFTP working locally, first set up a local user, with a username that you can use as a local SFTP server (e.g. localsftpuser).
```
TimboTambo, line 43: I think you still need to set a password which is used if you want to manually log in to the sftp server?

majikandy: you do actually need the password too when you run the line that adds it to authorised_keys

PR #671, opened 2020-04-16
Fix search client master CI build

sgrowe:
The master branch build for the search client is currently failing: https://github.com/MHRA/products/runs/589304928

Hopefully this fixes that


majikandy:
because this file is run on the master branch build only the only way to try this is here is to merge it

PR #672, opened 2020-04-16
Delete workflow should always result in a consistent state

m-doughty:
# Delete workflow should always result in a consistent state

The Doc Index Updater delete workflow currently can result in an inconsistent state where a result exists in storage but not in the search index. This PR will fix that.

Relates to #653.

### Acceptance Criteria

- [ ] The Doc Index Updater delete workflow always either resolves in both index & blob entry being deleted, or neither;
- [ ] If an index was found without a corresponding blob, this raises a NotFound error;

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/doc-index-updater/src/delete_manager/mod.rs
+++ b/medicines/doc-index-updater/src/delete_manager/mod.rs
@@ -100,6 +100,7 @@ async fn process_delete_message(
     search_client: impl Search + DeleteIndexEntry + CreateIndexEntry,
 ) -> Result<Uuid, ProcessMessageError> {
     let storage_container_name = std::env::var("STORAGE_CONTAINER").map_err(anyhow::Error::from)?;
+
     let index_record: IndexResult =
         get_index_record_from_content_id(message.document_content_id.clone(), &search_client)
             .await?;
@@ -161,6 +162,16 @@ pub async fn get_index_record_from_content_id(
         }
     }
     Err(ProcessMessageError::DocumentNotFoundInIndex(content_id))
+}
+
+pub async fn delete_from_index(
+    search_client: impl DeleteIndexEntry,
+    blob_name: &str,
+) -> Result<(), anyhow::Error> {
+    search_client
+        .delete_index_entry(&"metadata_storage_name".to_string(), &blob_name)
+        .await?;
+    Ok(())
 }
 
 #[cfg(test)]
--- a/medicines/doc-index-updater/src/delete_manager/mod.rs
+++ b/medicines/doc-index-updater/src/delete_manager/mod.rs
@@ -108,17 +108,16 @@ pub async fn process_message(message: DeleteMessage) -> Result<Uuid, ProcessMess
     let search_client = search_client::factory();
     let storage_client = storage_client::factory()
         .map_err(|e| anyhow!("Couldn't create storage client: {:?}", e))?;
-
-    process_delete_message(message, storage_client, search_client).await
+    let storage_container_name = std::env::var("STORAGE_CONTAINER").map_err(anyhow::Error::from)?;
+    process_delete_message(message, storage_client, search_client, storage_container_name).await
 }
 
 async fn process_delete_message(
     message: DeleteMessage,
     mut storage_client: impl DeleteBlob,
     search_client: impl Search + DeleteIndexEntry + CreateIndexEntry,
+    storage_container_name: String,
 ) -> Result<Uuid, ProcessMessageError> {
-    let storage_container_name = std::env::var("STORAGE_CONTAINER").map_err(anyhow::Error::from)?;
-
     let index_record: IndexResult =
         get_index_record_from_content_id(message.document_content_id.clone(), &search_client)
             .await?;
@@ -361,11 +360,12 @@ mod test {
         let removeable_message = given_we_have_a_delete_message().message;
         let search_client = given_a_search_client_that_returns_results();
         let storage_client = given_a_storage_client();
-
+        let storage_container_name = "storage_container_name".to_string();
         let result = block_on(process_delete_message(
             removeable_message,
             storage_client,
             search_client,
+            storage_container_name,
         ));
 
         assert_eq!(result.is_err(), false);
@@ -376,11 +376,13 @@ mod test {
         let removeable_message = given_we_have_a_delete_message().message;
         let search_client = given_a_search_client_that_returns_results();
         let storage_client = given_a_storage_client_that_cannot_delete_blob();
+        let storage_container_name = "storage_container_name".to_string();
 
         let result = block_on(process_delete_message(
             removeable_message,
             storage_client,
             search_client,
+            storage_container_name,
         ));
 
         match result {
@@ -403,11 +405,13 @@ mod test {
         let removeable_message = given_we_have_a_delete_message().message;
         let search_client = given_a_search_client_that_cannot_restore_index();
         let storage_client = given_a_storage_client_that_cannot_delete_blob();
+        let storage_container_name = "storage_container_name".to_string();
 
         let result = block_on(process_delete_message(
             removeable_message,
             storage_client,
             search_client,
+            storage_container_name,
         ));
 
         match result {
@@ -430,11 +434,13 @@ mod test {
         let removeable_message = given_we_have_a_delete_message().message;
         let search_client = given_a_search_client_that_cannot_delete_index();
         let storage_client = given_a_storage_client();
+        let storage_container_name = "storage_container_name".to_string();
 
         let result = block_on(process_delete_message(
             removeable_message,
             storage_client,
             search_client,
+            storage_container_name,
         ));
 
         match result {
@@ -665,8 +671,8 @@ mod test {
             match self.can_delete_blob {
                 true => Ok(()),
                 false => Err(AzureError::GenericErrorWithText(
-                            "blob could not be deleted".to_string(),
-                        ))
+                    "blob could not be deleted".to_string(),
+                )),
             }
         }
     }
--- a/medicines/doc-index-updater/src/delete_manager/mod.rs
+++ b/medicines/doc-index-updater/src/delete_manager/mod.rs
@@ -130,14 +130,11 @@ async fn process_delete_message(
             .create_index_entry(IndexEntry::from(index_record.clone()))
             .await
             .map_err(|err| {
-                ProcessMessageError::FailedRestoringIndex(
-                    blob_name.clone(),
-                    err.to_string()
-                )
+                ProcessMessageError::FailedRestoringIndex(blob_name.clone(), err.to_string())
             })?;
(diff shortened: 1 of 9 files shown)


medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -405,14 +408,14 @@ mod test {
         let removeable_message = given_we_have_a_delete_message().message;
         let search_client = given_a_search_client_that_cannot_restore_index();
         let storage_client = given_a_storage_client_that_cannot_delete_blob();
-        let storage_container_name = "storage_container_name".to_string();
+        given_the_necessary_env_vars_are_initialised();
```
majikandy, line 411: You can only run this test now with `make test` which makes it a bit less flexible as the run test button in vscode won't work for it. 

Is there any way around this? Is it totally necessary to use this env variable in the test, or more accurately, could this be injected into the function under test instead of hard wired to the env variable ?

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -247,6 +278,193 @@ mod test {
...
+            },
+        );
+    }
+
+    #[test]
+    fn index_with_doc_returns_success() {
+        let removeable_message = given_we_have_a_delete_message().message;
+        let search_client = given_a_search_client_that_returns_results();
+        let storage_client = given_a_storage_client();
+        given_the_necessary_env_vars_are_initialised();
```
majikandy, line 365: You can only run this test now with make test which makes it a bit less flexible as the run test button in vscode won't work for it.

Is there any way around this? Is it totally necessary to use this env variable in the test, or more accurately, could this be injected into the function under test instead of hard wired to the env variable ?

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -247,6 +278,193 @@ mod test {
...
+            &mut removeable_message,
+            error,
+            &state_manager,
+        ))
+        .unwrap();
+
+        assert_eq!(
+            state_manager.get_most_recently_set_status(),
+            JobStatus::Error {
+                message: String::from("Cannot restore index for blob Blob Id: Error message"),
```
rjkerrison, line 354: not immediately clear where `"Blob Id"` is coming from – might be worth passing a generated id into `given_failure_to_restore_index`

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -247,6 +278,193 @@ mod test {
...
+            }
+        }
+    }
+
+    fn given_the_necessary_env_vars_are_initialised() {
+        env::set_var("STORAGE_CONTAINER", "storage_container");
+    }
+
+    fn unset_process_delete_message_env_vars() {
+        env::set_var("STORAGE_CONTAINER", "");
```
rjkerrison, line 465: This doesn't do what it says. If we want to unset, can we use [`remove_var`](https://doc.rust-lang.org/std/env/fn.remove_var.html)?

medicines/doc-index-updater/src/delete_manager/mod.rs
TimboTambo: `given_the_necessary_env_vars_are_initialised` initialises the env vars necessary for this test to run, so removes manual prerequisite set-up. We're currently not using a testing framework that allows for set-up and tear-down, which would be a nice-to-have to make these steps cleaner. We could manually add a test wrapper or look at available testing crates as future steps.

medicines/doc-index-updater/src/service_bus_client/mod.rs
```diff
@@ -102,6 +102,10 @@ pub enum ProcessMessageError {
     SftpError(#[from] SftpError),
     #[error("Cannot find document with ID {0}")]
     DocumentNotFoundInIndex(String),
+    #[error("Cannot delete blob with ID {0}: {1}")]
+    FailedDeletingBlob(String, String),
+    #[error("Cannot restore index for blob {0}: {1}")]
```
rjkerrison, line 107: Seems odd that one error message says `"blob with ID <id>"` and the other says `"blob <id>"`. I'd go for the latter in both.

medicines/doc-index-updater/src/storage_client/mod.rs
```diff
@@ -1,15 +1,54 @@
-use azure_sdk_core::errors::AzureError;
-use azure_sdk_storage_core::prelude::Client;
+use async_trait::async_trait;
+use azure_sdk_core::{errors::AzureError, prelude::*, DeleteSnapshotsMethod};
+use azure_sdk_storage_blob::Blob;
+use azure_sdk_storage_core::prelude::*;
```
rjkerrison, line 4: do we have to have `prelude::*` or can we be explicit with `Client` etc? It's not clear when we use `Client` from where we get it

medicines/doc-index-updater/src/storage_client/mod.rs
```diff
@@ -1,15 +1,54 @@
...
+        blob_name: &str,
+    ) -> Result<(), AzureError> {
+        self.azure_client
+            .delete_blob()
+            .with_container_name(&container_name)
+            .with_blob_name(&blob_name)
+            .with_delete_snapshots_method(DeleteSnapshotsMethod::Include)
+            .finalize()
+            .await?;
+        Ok(())
```
rjkerrison, line 39: is there no value in returning the `DeleteBlobResponse`?

medicines/search-client/src/models.rs
```diff
@@ -1,14 +1,16 @@
...
+#[derive(Clone, Debug, Deserialize)]
 pub struct AzureHighlight {
     #[serde(rename = "content")]
     content: Vec<String>,
 }
 
-#[derive(Debug, Deserialize)]
-pub struct AzureResult {
+#[derive(Clone, Debug, Deserialize)]
+pub struct IndexResult {
```
rjkerrison, line 13: not sure there's such a big benefit in avoiding the `Azure` name if `highlights` has an `AzureHighlight` type and all the fields are married to Azure's naming

medicines/search-client/src/models.rs
```diff
@@ -80,3 +91,48 @@ pub struct IndexEntry {
     pub substance_name: Vec<String>,
     pub facets: Vec<String>,
 }
+
+// The AzureResult model does not contain all of the information we want in the index,
```
rjkerrison, line 95: comment is out-of-date with the `AzureResult` -> `IndexResult` rename

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -60,23 +62,41 @@ where
...
+                .await?;
+            let _remove = removeable_message.remove().await?;
+        }
+        ProcessMessageError::FailedRestoringIndex(_, _) => {
+            tracing::error!("{}", error.to_string());
+            state_manager
+                .set_status(
+                    removeable_message.get_message().job_id,
+                    JobStatus::Error {
+                        message: error.to_string(),
```
craiga, line 89: Why not use `error_message` instead of `error.to_string()`?

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -60,23 +62,41 @@ where
...
+                        code: "".to_string(),
+                    },
+                )
+                .await?;
+            let _remove = removeable_message.remove().await?;
```
craiga, line 94: Just to get this clear in my head, to get here, we've:

* found blob in index
* deleted from index
* failed to delete blob
* failed to re-add to index

Right?

If so, would it be worth writing a log message here to identify which blob caused the problem so it can be investigated/manually deleted? The blob name is already the first argument to this arm.

medicines/doc-index-updater/src/delete_manager/mod.rs
```diff
@@ -60,23 +62,41 @@ where
...
+        }
+        ProcessMessageError::FailedDeletingBlob(_, _) => {
+            tracing::error!("{}", error.to_string());
```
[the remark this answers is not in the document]
TimboTambo: I think it's important for this function to know the resulting actions to take for certain errors. Did you mean to make what is recoverable and not explicitly by naming the errors like e.g. `RecoverableError`?

medicines/doc-index-updater/src/delete_manager/mod.rs
TimboTambo: Valid concern! The `tracing::error` on line 84 should include the blob ID, since when this error is written to string, the blob name is included: `Cannot restore index for blob with ID <blob_id>: <wrapped error message>`

medicines/doc-index-updater/src/delete_manager/mod.rs
TimboTambo: True! Fixed.

medicines/search-client/src/models.rs
TimboTambo: Updated.

medicines/search-client/src/models.rs
TimboTambo: We do at least control the fields in `IndexResult` and `IndexEntry` since we define the shape of the index. The `AzureHighlight` I think is dictacted by Azure.

medicines/doc-index-updater/src/storage_client/mod.rs
TimboTambo: Not currently.

medicines/doc-index-updater/src/service_bus_client/mod.rs
TimboTambo: True - added.

medicines/doc-index-updater/src/delete_manager/mod.rs
TimboTambo: Set to use `remove_var`

medicines/doc-index-updater/src/delete_manager/mod.rs
TimboTambo: True - added

medicines/doc-index-updater/src/delete_manager/mod.rs
TimboTambo: I had to remove this unsetting in the end as it seems like tests are multi-threaded so this currently sometimes unsets it for other tests that rely on it. We'll have to come up with a better way of making it thread safe, perhaps with test suite-level set-up and tear-down. I'll create a ticket to address this.

medicines/doc-index-updater/src/delete_manager/mod.rs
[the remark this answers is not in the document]
rjkerrison: Yeah, that would make sense, if the error itself suggested that we should retry, or retrying could be the default and the unrecoverable errors could be `enum UnrecoverableError`.

PR #675, opened 2020-04-17
Add CORS support to API

craiga:
Add CORS support to API.

Required for #401, #403, #620, and #621.

### Acceptance Criteria

- [ ] Any web page can send a POST request to the API.

Note that _any_ web page should be permitted—this is public data!

### DIFF
--- a/medicines/api/src/main.rs
+++ b/medicines/api/src/main.rs
@@ -42,8 +42,7 @@ async fn healthz() -> impl actix_web::Responder {
 fn cors_middleware() -> actix_cors::CorsFactory {
     Cors::new()
         .allowed_methods(vec!["POST"])
-        .allowed_headers(vec![http::header::AUTHORIZATION, http::header::ACCEPT])
-        .allowed_header(http::header::CONTENT_TYPE)
+        .allowed_headers(vec![http::header::AUTHORIZATION, http::header::ACCEPT, http::header::CONTENT_TYPE])
         .max_age(3600)
         .finish()
 }
--- a/medicines/api/src/main.rs
+++ b/medicines/api/src/main.rs
@@ -1,5 +1,6 @@
 const PORT: u16 = 8000;
-use actix_web::{middleware, web, App, Error, HttpResponse, HttpServer};
+use actix_cors::Cors;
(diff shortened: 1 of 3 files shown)


StuartHarris:
We can add CORS support in the Istio Virtual Service so we don't need to add it to the web server itself. (https://istio.io/docs/reference/config/networking/virtual-service/#CorsPolicy)

craiga:
@StuartHarris Is there an advantage of doing that at the Istio level? Having to set up Istio sounds like it'd make development against a local copy of the API more difficult.

medicines/api/src/main.rs
```diff
@@ -38,9 +39,17 @@ async fn healthz() -> impl actix_web::Responder {
     "OK"
 }
 
+fn cors_middleware() -> actix_cors::CorsFactory {
+    Cors::new()
+        .allowed_methods(vec!["POST"])
+        .allowed_headers(vec![http::header::AUTHORIZATION, http::header::ACCEPT])
+        .allowed_header(http::header::CONTENT_TYPE)
+        .max_age(3600)
```
sgrowe, line 47: Do you not need to specify the allowed origins as well?

rjkerrison:
If we're spinning up in kubernetes locally (not unreasonable, but we're not there yet) then we can make the Istio change. Until then, we need a way to allow CORS on at least localhost. Maybe this can be feature-toggled when there's an alternative.

StuartHarris:
Ahh cool! Yeah, there's no harm in having it in both places. When it's running the pre-flight requests won't get as far as the pod, but that's fine :-)

StuartHarris:
@craiga yeah, you're right, sorry, it would be silly to have to run it under Istio locally. In production, though, it makes sense to do it in the Virtual Service in the spirit of moving the cross-cutting concerns from the service to the sidecar.

medicines/api/src/main.rs
craiga: @sgrowe We deliberately decided to not set allowed origins as this is a public API. One of the acceptance criteria is that it can be used by third party developers.

I'll look into whether we can combine `allowed_headers` and `allowed_header`.

PR #681, opened 2020-04-20
Substance page uses API

rjkerrison:
# Points the substance page to the API

Substance page using API. Hidden by feature switch for now, see #693.

### Acceptance Criteria

- [ ] Substance page uses API, passing a substance name and receiving a list of products
- [ ] Caching of requests, so users don't fire the same graphql query twice in a session

### Testing information

1. Run `make` in the `medicines/api` directory.
2. Run `yarn dev` in the `medicines/web` directory.
3. Navigate to `localhost:3000` and go to a substance page via the index.
4. See the request in your browser's network dev tools. Notice that it's a graphql request to `localhost:8000`.

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


medicines/api/src/product.rs
```diff
@@ -33,14 +34,14 @@ pub fn handle_doc(document: &IndexResult, products: &mut Vec<Product>) {
...
 }
 
-pub async fn get_products_by_substance_name(
-    substance_name: &str,
+pub async fn get_substance_with_products(
+    name: &str,
     client: &impl Search,
-) -> Vec<Product> {
+) -> Substance {
     // Get a list of documents from Azure which are about products containing the
```
majikandy, line 41: Is this comment displayed somewhere?  if not then it doesn't appear to add anymore information than the function name itself - other than saying Azure

medicines/api/src/schema.rs
```diff
@@ -2,25 +2,25 @@ use juniper::{FieldResult, RootNode};
...
 
 #[juniper::graphql_object(Context = AzureContext)]
 impl QueryRoot {
-    async fn products(
+    async fn substance(
         context: &AzureContext,
-        substance_name: Option<String>,
-    ) -> FieldResult<Vec<Product>> {
-        match substance_name {
+        name: Option<String>,
```
majikandy, line 15: is it possible for this to not be optional and therefore not have a need to handle the None arm or is this because you  could make the call from the client this way so you want to be explicit?

medicines/api/src/substance.rs
```diff
@@ -1,12 +1,17 @@
...
     name: String,
+    products: Option<Vec<Product>>,
 }
 
 impl Substance {
+    pub fn new(name: String, products: Option<Vec<Product>>) -> Self {
+        Self { name, products }
+    }
+
     #[allow(dead_code)]
```
majikandy, line 15: is this needed?

medicines/api/src/substance.rs
```diff
@@ -22,6 +27,7 @@ pub async fn get_substances(first: i32) -> Substances {
         .take(first as usize)
         .map(|x| Substance {
             name: x.to_owned().to_owned(),
```
majikandy, line 29: double `to_owned`?

medicines/api/src/substance.rs
```diff
@@ -22,6 +27,7 @@ pub async fn get_substances(first: i32) -> Substances {
         .take(first as usize)
```
majikandy, line 27: Is this stubbed out 1000 ibuprofens something that needs replacing before this goes to prod?

medicines/web/Makefile
```diff
@@ -16,3 +16,7 @@ set-env: ## Takes your current .env file and replaces the keyvault value with th
 	--vault-name mhra-dev \
 	--name web-env \
 	--file .env
+
+.PHONY: cypress
+cypress: ## Launch Cypress
+	yarn run cypress open --env $$(cat .env 2> /dev/null | tr "\n", "," | sed 's/,$$/ /g')
```
majikandy, line 22: one error when running cypress tests...

```
Navigate to Paracetamol via A-Z indexfailed
CypressError: cy.route() must be called with a url. It can be a string or regular expression.
```

medicines/api/src/product.rs
rjkerrison: I agree, it's pointless if it's not docs

medicines/api/src/substance.rs
rjkerrison: as part of #403

medicines/api/src/substance.rs
rjkerrison: an unfortunate consequence of the library we're using not being smart enough to recognise that this method is used

medicines/api/src/schema.rs
rjkerrison: if this isn't `Option`, then it becomes a required field in the GraphQL

medicines/api/src/product.rs
```diff
@@ -33,26 +34,19 @@ pub fn handle_doc(document: &IndexResult, products: &mut Vec<Product>) {
     }
 }
 
-pub async fn get_products_by_substance_name(
-    substance_name: &str,
-    client: &impl Search,
-) -> Result<Vec<Product>, reqwest::Error> {
-    // Get a list of documents from Azure which are about products containing the
-    // substance name.
+pub async fn get_substance_with_products(name: &str, client: &impl Search) -> Result<Substance, reqwest::Error> {
```
TimboTambo, line 37: I think it'd be good to stick with `substance_name` here, just makes it that bit more explicit since both products and substances are in play here.

PR #684, opened 2020-04-21
Put API key in header in search client

sgrowe:
This now contains just a bug fix for the search client, where it was previously putting the `AZURE_API_ADMIN_KEY` env variable in the url which is not allowed for admin keys. 

Really the API only needs a query key not an admin key anyway as it doesn't do any creating or deleting. But as the search client expects a key called `AZURE_API_ADMIN_KEY` so I thought it best to just give it an admin key for the time being to avoid confusion in future and we can maybe refactor it later if we want


rjkerrison:
@craiga since you merged this branch into #681, do we need this PR? I was expecting us to merge this one into master ahead of #681 so we can see the API deployment work before web is targeting it

craiga:
@rjkerrison What you say makes sense. I wanted to merge this into #681 to make sure it's obvious this is required by that PR, and to stay on top of the inevitable conflicts that would arise.

TimboTambo:
@sgrowe Can you rename this PR to reflect the changes it's actually bringing in now, along with some further description and a link to the related ticket?

sgrowe:
@TimboTambo sure, it looks like this PR only contains one bug fix now so I've updated the ticket description for that

PR #688, opened 2020-04-22
Remove final blank line with make get-env

majikandy:

### DIFF
--- a/medicines/api/Makefile
+++ b/medicines/api/Makefile
@@ -17,7 +17,7 @@ get-env: ## Gets the environment variables from azure keyvault into .env file
 	  --vault-name mhra-dev \
 	  --name api-env \
 	  --query value \
-	  --output tsv > .env
+	  --output tsv | sed '$$d' > .env
 
 .PHONY: set-env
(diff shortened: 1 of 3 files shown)


medicines/api/Makefile
```diff
@@ -17,7 +17,7 @@ get-env: ## Gets the environment variables from azure keyvault into .env file
 	  --vault-name mhra-dev \
 	  --name api-env \
 	  --query value \
-	  --output tsv > .env
+	  --output tsv | sed '$$d' > .env
```
TimboTambo, line 20: Won't this remove the last line, regardless of what's on it? It looks like `sed '${/^$/d;}'` would delete the last line only if it's empty.

majikandy: Yes this was very specific due to the way the to and from kv seems to always introduce a blank line at the end.

If we are comfortable with removing all blank lines we could do:-
`sed '/^$/d'`

TimboTambo: I'd be happy with that - blank lines, what are they good for? Absolutely nothing (in auto-generated files).

majikandy: hahahaha

PR #695, opened 2020-04-22
Get username and password from env for test

majikandy:
Rather than using username:password - get it from env variables

Also added `make e2e` to be able to run the ignored with the env variables set.

### DIFF
--- a/medicines/doc-index-updater/Makefile
+++ b/medicines/doc-index-updater/Makefile
@@ -10,6 +10,10 @@ default: ## Run locally
 test: ## Run tests [TEST=test_name (optional)]
 	export $$(cat .env .env.overrides 2> /dev/null | xargs) && cargo test $$TEST
 
+.PHONY: e2e
+e2e: ## Run end to end acceptance tests [TEST=test_name (optional)]
+	export $$(cat .env .env.overrides 2> /dev/null | xargs) && cargo test $$TEST -- --ignored
+
 .PHONY: get-env
 get-env: ## Gets the environment variables from azure keyvault into .env file
 	az keyvault secret show \
--- a/medicines/doc-index-updater/tests/support/document_api.rs
+++ b/medicines/doc-index-updater/tests/support/document_api.rs
@@ -8,7 +8,7 @@ use doc_index_updater::{
 use reqwest::Error;
 use uuid::Uuid;
 
-fn get_basic_auth_credentials() -> BasicAuth {
+fn get_basic_auth_credentials() -> (String, Option<String>) {
     let username = get_env("BASIC_AUTH_USERNAME").unwrap();
     let password = if let Ok(p) = get_env("BASIC_AUTH_PASSWORD") {
         Some(p)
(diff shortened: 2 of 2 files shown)


medicines/doc-index-updater/tests/support/document_api.rs
```diff
@@ -1,15 +1,38 @@
...
 
+pub fn get_basic_auth_credentials() -> (String, String) {
+    let username: String = get_env("BASIC_AUTH_USERNAME").unwrap();
+    let password: String = get_env("BASIC_AUTH_PASSWORD").unwrap();
+
+    (username, password)
+}
+
+pub trait BasicAuthFromEnv {
+    fn with_auth(self) -> RequestBuilder;
```
rjkerrison, line 19: trait `WithAuth` to match `with_auth` and return type `Self` – that it returns `RequestBuilder` is an implementation detail for `RequestBuilder`

medicines/doc-index-updater/tests/support/document_api.rs
```diff
@@ -1,15 +1,38 @@
...
+}
+
+impl BasicAuthFromEnv for RequestBuilder {
+    fn with_auth(self) -> RequestBuilder {
+        let (username, password) = get_basic_auth_credentials();
```
rjkerrison, line 24: if you did want this trait to imply basic auth, it could be `WithBasicAuth / with_basic_auth(self, String, String)`, with `username` and `password` passed as arguments

majikandy: It's more like `WithBasicAuthFromEnvironmentVars` and `with_basic_auth_from_environment_vars`

medicines/doc-index-updater/Makefile
```diff
@@ -10,6 +10,10 @@ default: ## Run locally
 test: ## Run tests [TEST=test_name (optional)]
 	export $$(cat .env .env.overrides 2> /dev/null | xargs) && cargo test $$TEST
 
+.PHONY: e2e
+e2e: ## Run end to end acceptance tests [TEST=test_name (optional)]
+	export $$(cat .env .env.overrides 2> /dev/null | xargs) && cargo test $$TEST -- --ignored
```
sgrowe, line 15: Doesn't seem great overloading ignored tests as a organising tests. Could we split the e2e tests out into a separate crate instead?

majikandy: I agree but that wasn't part of this change - actually the indirection by having e2e target here allows us to change that.

medicines/doc-index-updater/tests/support/document_api.rs
```diff
@@ -15,12 +15,12 @@ pub fn get_basic_auth_credentials() -> (String, String) {
     (username, password)
 }
 
-pub trait BasicAuthFromEnv {
-    fn with_auth(self) -> RequestBuilder;
+pub trait WithBasicAuthFromEnvironmentVars {
+    fn with_basic_auth_from_environment_vars(self) -> RequestBuilder;
```
rjkerrison, line 19: trait still has `RequestBuilder` return type

PR #697, opened 2020-04-22
Use sccache to speed up CI builds

sgrowe:
Uses [sccache](https://github.com/mozilla/sccache) as part of the docker build so that the compiled rust code is automatically cached in Azure blob storage. This speeds up CI builds a fair bit as currently the docker build & rust compile steps aren't cached at all, so even if the dependencies of our projects don't change they are all compiled from scratch on every build.

This does add some complexity to the docker build but there's a decent reduction in build times. It seems to reduce the build time for `medicines/api` [from ~13mins to ~6mins](https://github.com/MHRA/products/actions?query=workflow%3Amedicines-api-branch) and for doc-index-updater from [~11mins to ~8mins](https://github.com/MHRA/products/actions?query=workflow%3Adoc-index-updater-branch).


sgrowe:
> how possible is this to use for local builds ?

@majikandy it should be pretty straightforward.

If you run `make docker-build` locally this will use sccache already, but if you want to use it outside of docker then you need to:
- [install sccache](https://github.com/mozilla/sccache#installation)
- run `make get-env` to download `.env.build` which contains `SCCACHE_AZURE_CONNECTION_STRING`
- set these environment variables:
  - `RUSTC_WRAPPER` - path to sccache
  - `SCCACHE_AZURE_CONNECTION_STRING` - same value as in `.env.build`
  - `SCCACHE_AZURE_BLOB_CONTAINER=cicache`

I'm not sure how much benefit there is in setting it up outside of docker though as `cargo build` automatically caches stuff to your local filesystem anyway. It would only really be if someone else has compiled a new dependency before you that it would help really.

medicines/api/Cargo.toml
```diff
@@ -14,10 +14,10 @@ actix-cors = "0.2.0"
 futures = "0.3.1"
 juniper = { git = "https://github.com/graphql-rust/juniper", features = ["async"], rev="211737cee0c6b79ad86246d69ed39e2725fc2110" }
 listenfd = "0.3.3"
-reqwest = { version = "0.10.1", features = ["json"] }
+reqwest = { version = "0.10.4", features = ["json"] }
```
rjkerrison, line 17: are these cargo updates related to enabling sccache?

medicines/api/Cargo.toml
```diff
@@ -14,10 +14,10 @@ actix-cors = "0.2.0"
...
 search_client =  { path = "../search-client" }
 tracing = "0.1.13"
-tracing-subscriber = "0.2.4"
-serde = "1.0.103"
-serde_derive = "1.0.103"
-serde_json = "1.0.44"
+tracing-subscriber = "0.2.5"
+serde = "1.0.106"
+serde_derive = "1.0.106"
+serde_json = "1.0.51"
```
rjkerrison, line 23: same here, are these cargo updates related to enabling sccache?

medicines/api/Cargo.toml
sgrowe: No, I just wanted to trigger a full rebuild and figured a minor version update was probably a good thing to do anyway

rjkerrison: probably okay if it builds - was worried it was causing unnecessary conflicts but if they're resolved it's no problem

PR #703, opened 2020-04-23
Drug index page uses API

sgrowe:
Updates the drug index page to use the `medicines/api` project instead of calling the Azure search API directly


medicines/web/src/pages/substance/index.tsx
```diff
@@ -11,7 +11,7 @@ import {
 import { useLocalStorage } from '../../hooks';
 import { IProduct } from '../../model/substance';
 import Events from '../../services/events';
-import { products } from '../../services/graphql-loader';
+import { products } from '../../services/products-loader';
```
rjkerrison, line 14: @sgrowe I was expecting that the `graphql-loader` would expand a little, so that we can make one big request to the graphql endpoint like `substance-loader` does to Azure, to get a big drill-down

.github/workflows/medicines-web-branch.yaml
```diff
@@ -48,6 +48,16 @@ jobs:
...
 
+      - name: Write variables to .env
+        working-directory: medicines/web
+        run: |
+          {
+            echo "AZURE_SEARCH_SERVICE=\"$AZURE_SEARCH_SERVICE\""
+            echo "AZURE_SEARCH_INDEX=\"$AZURE_SEARCH_INDEX\""
+            echo "AZURE_SEARCH_KEY=\"$AZURE_SEARCH_KEY\""
+            echo "GRAPHQL_URL=\"$GRAPHQL_URL\""
+          } > .env
```
rjkerrison, line 59: is this because `yarn` doesn't read from env but from `.env`?

medicines/web/src/pages/substance/index.tsx
sgrowe: Ah ok. I've just tried to keep an identical API to the existing substances loader so that switching between them with the feature flag is easy. 

I guess once the flag is removed that might be a good time to add some tool for combining graphql requests though?

medicines/web/src/services/products-loader.ts
```diff
@@ -0,0 +1,30 @@
...
+      name
+      documentCount
+    }
+  }
+}`;
+
+  const variables = { substanceName };
+
+  const resp = await graphqlRequest<
+    { substance: { products: IProduct[] } },
```
rjkerrison, line 19: this is awesome

.github/workflows/medicines-web-branch.yaml
sgrowe: I updated the `yarn test-e2e` script to read from `.env` just as a convenience for people so I had to add this so that `.env` exists when that script is called in CI

PR #706, opened 2020-04-23
added warp recovery for correct status codes

### LINKED ISSUE #700: Unauthorised request returns wrong error code
**Describe the bug**
An unauthorized request to the document-index-updater returns a 500 (internal server error) but should return a 401 (unauthorized). This makes it more difficult for clients to diagnose possible issues with requests, as it surfaces like an issue on the document-index-updater side.

**To Reproduce**

1. Perform `curl -L -X GET 'https://doc-index-updater.test.mhra.gov.uk/jobs/f9ec57a6-3d54-43d6-a1ee-924117f2b72b' \
-H 'Content-Type: application/json' \
-H 'Authorization: Basic dXNrcm5hbWU6cGFzc3dvcmQ='`
2. Observe 500 error response

**Expected behavior**
401 response

**Screenshots**
N/A

StuartHarris:
Adds a warp recovery handler to transform custom rejections into the correct HTTP Status Codes.

fixes #700

### DIFF
--- a/medicines/doc-index-updater/src/main.rs
+++ b/medicines/doc-index-updater/src/main.rs
@@ -1,10 +1,11 @@
 use doc_index_updater::{
-    create_manager, delete_manager, document_manager, get_env_or_default, health, state_manager,
+    auth_manager::AuthenticationFailed, create_manager, delete_manager, document_manager,
+    get_env_or_default, health, state_manager,
 };
 use state_manager::get_client;
-use std::{error, net::SocketAddr, time::Duration};
+use std::{convert::Infallible, error, net::SocketAddr, time::Duration};
 use tracing::Level;
-use warp::Filter;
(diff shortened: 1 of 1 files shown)


medicines/doc-index-updater/src/main.rs
```diff
@@ -91,6 +93,41 @@ fn create_redis_url(server: String, port: String, key: String) -> String {
...
+        code = StatusCode::INTERNAL_SERVER_ERROR;
+        message = "UNHANDLED_REJECTION";
+    }
+
+    let json = warp::reply::json(&ErrorMessage {
+        code: code.as_u16(),
+        message: message.into(),
+    });
+
+    Ok(warp::reply::with_status(json, code))
```
TimboTambo, line 128: Warp will automatically log the outgoing response, so we don't need to manually log here, right?

PR #710, opened 2020-04-23
Search client page size restriction parameters are misnamed.

craiga:
Removing them to reduce confusion.

medicines-api-branch

doc-index-updater-branch

Idris smiling

### DIFF
--- a/medicines/search-client/src/lib.rs
+++ b/medicines/search-client/src/lib.rs
@@ -196,9 +196,6 @@ fn build_search(
             ("api-version", config.api_version.as_str()),
             ("highlight", "content"),
             ("queryType", "full"),
-            ("@count", "true"),
-            ("@top", "10"),
-            ("@skip", "0"),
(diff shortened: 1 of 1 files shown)


medicines/search-client/src/lib.rs
```diff
@@ -196,9 +196,6 @@ fn build_search(
             ("api-version", config.api_version.as_str()),
             ("highlight", "content"),
             ("queryType", "full"),
-            ("@count", "true"),
```
rjkerrison, line 199: is it better to fix them now? `$` it up?

craiga: I _think_ I remember reading that including count, top and skip increases query complexity, which means we'll hit usage limits sooner. [The documentation is very vague about this though.](https://docs.microsoft.com/en-us/azure/search/search-limits-quotas-capacity#queries-per-second-qps)

PR #712, opened 2020-04-24
Query substances by first letter

sgrowe:
Adds ability to query for substances by the first letter in their name. Needed for #403


medicines/api/src/substance.rs
```diff
@@ -1,14 +1,16 @@
...
+use search_client::{models::IndexResults, Search};
+use std::collections::BTreeMap;
 
-#[derive(juniper::GraphQLObject)]
+#[derive(Debug, juniper::GraphQLObject)]
 #[graphql(description = "An active ingredient found in medical products")]
 pub struct Substance {
     name: String,
-    products: Option<Vec<Product>>,
+    products: Vec<Product>,
```
rjkerrison, line 9: I think by doing this you make `product` non-optional on Substance in the graphQL

medicines/api/src/substance.rs
```diff
@@ -1,31 +1,112 @@
...
+        start_cursor: edges
+            .first()
+            .map(|edge| edge.cursor.clone())
+            .unwrap_or(first_cursor.clone()),
+        end_cursor: edges
+            .last()
+            .map(|edge| edge.cursor.clone())
+            .unwrap_or(first_cursor),
+        has_previous_page: false,
+        has_next_page: false,
```
rjkerrison, line 92: if we're always sending the full list, do we need to return the pagination type, or could we just return `Vec<Substance>`?

medicines/api/src/substance.rs
```diff
@@ -1,31 +1,112 @@
...
+
+fn format_search_results(results: IndexResults, letter: char) -> Substances {
+    let mut substances: BTreeMap<&str, BTreeMap<&str, i32>> = BTreeMap::new();
+
+    let letter_string = letter.to_string();
+
+    results
+        .search_results
+        .iter()
+        .filter(|result| result.facets.first() == Some(&letter_string))
```
craiga, line 42: Is it possible to search for index results where facets start with a letter so we don't have to do this?

medicines/api/src/substance.rs
```diff
@@ -1,31 +1,112 @@
...
         Self { name, products }
     }
-
-    #[allow(dead_code)]
-    fn name(&self) -> &str {
-        &self.name
-    }
 }
 
 pagination! {Substances, SubstanceEdge, Substance}
```
craiga, line 19: There's no need to paginate substances, so we should get rid of this.

medicines/api/src/substance.rs
sgrowe: I'm not sure. I haven't checked but I assume not as the current JS code does this as well and I've just been trying to mimic that: https://github.com/MHRA/products/blob/master/medicines/web/src/services/substance-loader.ts#L11

sgrowe: Could be a good thing to investigate as an iterative improvement down the line?

PR #716, opened 2020-04-24
Implement pagination in the search client

craiga:
Pagination in the search client

### DIFF
--- a/medicines/search-client/src/lib.rs
+++ b/medicines/search-client/src/lib.rs
@@ -19,6 +19,11 @@ pub struct AzureSearchClient {
     config: AzureConfig,
 }
 
+pub struct AzurePagination {
+    pub result_count: u32,
+    pub offset: u32,
+}
+
 impl Default for AzureSearchClient {
     fn default() -> Self {
         Self::new()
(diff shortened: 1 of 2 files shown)


medicines/search-client/src/lib.rs
```diff
@@ -166,43 +192,70 @@ impl CreateIndexEntry for AzureSearchClient {
...
+    match (result_count, offset) {
+        (Some(result_count), Some(offset)) => Ok(request_builder
+            .query(&[
+                ("$top", result_count.to_string()),
+                ("$skip", offset.to_string()),
+            ])
+            .build()?),
+        (None, None) => Ok(request_builder.build()?),
+        _ => Err(anyhow::anyhow!(
+            "If one of result_count or offset is set, the other must be set as well."
```
rjkerrison, line 256: could they be unified into a `Pagination` type, then we pass in just `Option<Pagination>`? even `Option<(u32, u32>) should work

medicines/search-client/src/lib.rs
```diff
@@ -166,43 +192,70 @@ impl CreateIndexEntry for AzureSearchClient {
...
         .error_for_status()?
         .json::<IndexResults>()
         .await
+        .map_err(|e| anyhow::anyhow!(e))
 }
 
 fn build_search(
     search_term: &str,
+    result_count: Option<u32>,
+    offset: Option<u32>,
```
rjkerrison, line 222: could these be `i32` instead? I don't know what a negative integer would be useful for

craiga: They could be `i32`, but that would allow negative result count and offsets, which don't make any sense.

rjkerrison: right, think I just muddled up which integer is which, don't mind me

PR #717, opened 2020-04-24
Live document data in the API

craiga:
Hook up the document API with real, live data.

Example of the API output:

Screenshot 2020-04-27 at 17 08 52

### DIFF
--- a/medicines/api/src/document.rs
+++ b/medicines/api/src/document.rs
@@ -49,7 +49,7 @@ pub async fn get_documents(
             let bytes = base64::decode(after)?;
             let string = std::str::from_utf8(&bytes)?;
             string.parse::<i32>()? + 1
-        },
+        }
         None => 0,
     };
 
@@ -79,9 +79,10 @@ pub async fn get_documents(
         })
         .collect();
 
+    let result_count = azure_result.search_results.len() as i32;
     let total_count = azure_result.count.unwrap_or(0);
     let has_previous_page = offset != 0;
-    let has_next_page = offset + result_count <= total_count;
+    let has_next_page = offset + result_count < total_count;
     let start_cursor = base64::encode(offset.to_string());
     let end_cursor =
         base64::encode(std::cmp::min(total_count, offset + result_count - 1).to_string());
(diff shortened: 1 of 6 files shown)


medicines/api/src/document.rs
```diff
@@ -55,31 +39,57 @@ pagination! {Documents, DocumentEdge, Document}
...
+    let total_count = azure_result.count.unwrap_or(0);
+    let has_previous_page = offset != 0;
+    let has_next_page = offset + result_count <= total_count;
+    let start_cursor = offset.to_string();
+    let end_cursor = std::cmp::min(total_count, offset + result_count - 1).to_string();
+
+    Ok(Documents {
         edges,
-        total_count: 1000,
+        total_count: total_count,
```
rjkerrison, line 87: clippy is gonna tell you not to do this, just use `total_count`, same for four lines below

medicines/api/src/document.rs
```diff
@@ -55,31 +39,229 @@ pagination! {Documents, DocumentEdge, Document}
...
+    search: String,
     first: Option<i32>,
     last: Option<i32>,
     before: Option<String>,
     after: Option<String>,
-) -> Documents {
-    let placeholder_names: [&str; 1000] = ["A cool document"; 1000];
-    let edges = placeholder_names
+) -> Result<Documents, anyhow::Error> {
+    // TODO: Do something with last and after.
```
craiga, line 48: Need to do something about this. Perhaps they should be removed for now?

medicines/api/src/document.rs
```diff
@@ -55,31 +39,229 @@ pagination! {Documents, DocumentEdge, Document}
...
-        // .take(first as usize)
-        .map(|&name| Document::name_only(name.to_string()))
-        .map(|document| DocumentEdge {
-            node: document,
-            cursor: "cursor".to_owned(),
+        .map(|search_result| Document::from(search_result.clone()))
+        .map(|document| {
+            let edge = DocumentEdge {
+                node: document,
+                cursor: cursor.to_string(),
```
craiga, line 72: As discussed with @sgrowe, these should be opaque. For now, we'll base 64 encode the page numbers so that our client can still jump to any page.

medicines/api/src/document.rs
```diff
@@ -55,31 +39,229 @@ pagination! {Documents, DocumentEdge, Document}
...
+            return edge;
         })
         .collect();
 
-    Documents {
+    let total_count = azure_result.count.unwrap_or(0);
+    let has_previous_page = offset != 0;
+    let has_next_page = offset + result_count <= total_count;
+    let start_cursor = offset.to_string();
+    let end_cursor = std::cmp::min(total_count, offset + result_count - 1).to_string();
```
craiga, line 83: As mentioned previously, these should be opaque.

medicines/api/src/document.rs
craiga: From [the spec](https://relay.dev/graphql/connections.htm#sec-Arguments):

> A field that returns a Connection Type must include forward pagination arguments, backward pagination arguments, or both. These pagination arguments allow the client to slice the set of edges before it is returned.

To me, this says that we can remove `last` and `before` from this list.

rjkerrison: can't see much of a use case, so long as there's a `totalResults` for making the last page pagination links

medicines/search-client/src/lib.rs
```diff
@@ -20,8 +20,8 @@ pub struct AzureSearchClient {
 }
 
 pub struct AzurePagination {
-    pub result_count: u32,
-    pub offset: u32,
+    pub result_count: i32,
+    pub offset: i32,
```
sgrowe, line 24: Don't these both want to stay as `u32` so they can't ever be negative?

medicines/api/src/document.rs
```diff
@@ -55,31 +39,235 @@ pagination! {Documents, DocumentEdge, Document}
...
-        .map(|document| DocumentEdge {
-            node: document,
-            cursor: "cursor".to_owned(),
+        .map(|search_result| Document::from(search_result.clone()))
+        .map(|document| {
+            let edge = DocumentEdge {
+                node: document,
+                cursor: base64::encode(cursor.to_string()),
+            };
+            cursor += 1;
```
sgrowe, line 77: Just a suggestion but you could use [`.enumerate`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate) to get the index of the document. e.g.

```rust
.enumerate()
.map(|(i, document| {
  DocumentEdge {
    node: document,
    cursor: base64::encode((offset + i).to_string()),
  }
})

File: medicines/api/src/document.rs [hunk 7]
```diff
@@ -55,31 +39,235 @@ pagination! {Documents, DocumentEdge, Document}
 
 pub async fn get_documents(
     client: &impl Search,
-    search: Option<String>,
+    search: String,
     first: Option<i32>,
```
sgrowe, line 43: Would this be better off as a `u32` as you don't want to ever pass in a negative number?

medicines/search-client/src/lib.rs
craiga: Juniper had problems with dealing with `u32`s.

sgrowe: Ah, fair enough

PR #720, opened 2020-04-24
Split of non-prod and dev update-hosts script

majikandy:
Realised that dev also has host file entries so split them into 2 files


majikandy:
@sgrowe agree, done

infrastructure/scripts/update-hosts.sh
```diff
@@ -1,12 +1,49 @@
...
+#!/usr/bin/env bash
+set -euo pipefail
 
-sh ./update-kubernetes-config.sh
+update_hosts_file() {
+    if [ -z "$1" ]; then
+        echo 'setting postfix to empty'
+        ENV_POSTFIX=''
+    else
+        echo "setting postfix to $1"
```
sgrowe, line 9: Should these two `echo` lines be removed?

majikandy: ah good spot

PR #721, opened 2020-04-28
Docker BuildKit

StuartHarris:
Docker [`buildkit`](https://docs.docker.com/develop/develop-images/build_enhancements/) has new features (e.g. for secrets needed during the build), but also could be faster because it builds images in parallel where it can.

This PR is to test (and potentially use) this feature.

- [x] build with Docker BuildKit
- [x] use Rust 1.43 for building
- [x] tidy up some readme files en-route

### DIFF
--- a/hello-world/Cargo.lock
+++ b/hello-world/Cargo.lock
@@ -34,6 +34,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-sse"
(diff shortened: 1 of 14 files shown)


sgrowe:
Would be interesting to add this to the doc-index-updater & medicines api to see what impact it has as those have slightly beefier builds?

PR #722, opened 2020-04-28
Simplify alert query

majikandy:
simplified the kusto


majikandy:
Re WARN messages, not at the moment but we could create a different severity level alert later if we decided we wanted to

PR #726, opened 2020-04-28
Automatically create & destroy the dev cluster using github actions

sgrowe:
Saves MHRA money and saves us the trouble of manually creating and destroying the cluster everyday. There's still a manual step needed to apply the kubernetes configs to the dev cluster so I might add that later.

Another future enhancement we might want to make at a later date is to completely destroy the whole dev environment either everyday or over the weekend to save even more money. We'd then have to worry about updating the `.env` files and creating the search index though

### DIFF
--- a/.github/workflows/dev-cluster-create.yaml
+++ b/.github/workflows/dev-cluster-create.yaml
@@ -1,0 +1,50 @@
+name: Bring up dev cluster
+
+on:
+  schedule:
+    - 30 8 * * 1-5 # 8:30, Monday - Friday
+
+env:
+  TF_IN_AUTOMATION: "1"
+  TF_VAR_CLIENT_ID: ${{ secrets.TF_VAR_CLIENT_ID }}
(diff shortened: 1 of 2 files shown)


sgrowe:
I think this needs a bit more thought security wise. I've checked and anyone who is signed in to any github account can view the logs and artifacts for our github actions, and we don't want to accidentally expose access keys to a service bus or our azure subscription or something. 

I think the two main issues are:
- `terraform apply` writes all of our "outputs" to stdout. But this can be solved easily enough by marking them as [`sensitive = true`](https://www.terraform.io/docs/configuration/outputs.html#sensitive-suppressing-values-in-cli-output)
- the generated terraform plan probably contains a bunch of sensitive information. I like the idea of keeping a copy of the plan that was run so that if the automation messes up we can check why, but I'm not sure how best to store it securely

Maybe we should encrypt the terraform plan file..? Or store it on a private Azure blob storage? Or perhaps just redirect all of the terraform output to a file and store that on blob storage or something? Or use [terraform cloud](https://www.terraform.io/docs/cloud/overview.html)?

PR #727, opened 2020-04-28
Put CPD CDN profile in West Europe region

sgrowe:
You can't create CDN profiles in UK south sadly, which is where the rest of the non-prod infrastructure is located.

### DIFF
--- a/hello-world/Cargo.lock
+++ b/hello-world/Cargo.lock
@@ -34,6 +34,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "async-sse"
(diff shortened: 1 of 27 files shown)


StuartHarris:
Do you think we should move everything to West Europe then?

StuartHarris:
Closed by mistake!

sgrowe:
> Do you think we should move everything to West Europe then?

Yeah might as well, good shout

rjkerrison:
oh, I just did #857 which solves this in a slightly different way. just gonna close this now

PR #731, opened 2020-04-29
Multi level filtering

rjkerrison:
# Unlocking pagination within results for Product

Carries on with #401 – gives us pagination! If you request documents for a product, you can optionally request fewer than the full result set. You can even request to skip some!

Damn fine filtering

### Acceptance Criteria

- [ ] Introduces the ability to request a certain number of documents for a product

### Testing information

Run `make`. Go to http://localhost:8000/graphiql.

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/api/src/document.rs
+++ b/medicines/api/src/document.rs
@@ -37,6 +37,16 @@ impl From<IndexResult> for Document {
 
 pagination! {Documents, DocumentEdge, Document}
 
+pub fn get_document_edges(docs: Vec<Document>, offset: i32) -> Vec<DocumentEdge> {
+    docs.into_iter()
+        .enumerate()
+        .map(|(i, document)| DocumentEdge {
+            node: document,
+            cursor: base64::encode((i as i32 + offset).to_string()),
+        })
+        .collect()
+}
+
 pub async fn get_documents(
     client: &impl Search,
     search: String,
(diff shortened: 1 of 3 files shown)


medicines/api/src/product.rs
```diff
@@ -25,6 +26,28 @@ impl Product {
...
+        let docs = self.documents.clone().into_iter();
+        let docs = match first {
+            Some(t) => docs.take(t as usize).collect(),
+            None => docs.collect(),
+        };
+
+        let offset = match skip {
+            Some(a) => a,
+            None => 0,
+        };
```
sgrowe, line 45: It's perhaps trying to be a bit too brief but you could do `skip.unwrap_or_default()` or `skip.unwrap_or(0)` instead?

rjkerrison: yeah, either is fine really, I've just tried to get away from `unwrap_or` because it's enforcing a bad habit of writing `unwrap` in production code

medicines/api/src/document.rs
```diff
@@ -37,6 +37,45 @@ impl From<IndexResult> for Document {
 
 pagination! {Documents, DocumentEdge, Document}
 
+fn get_document_edges(docs: Vec<Document>, offset: i32) -> Vec<DocumentEdge> {
```
craiga, line 40: Love this approach.

Would `build_` be a better prefix than `get_`? To me, "get" implies retrieving from some data source, whereas "build" would be about building structures.

medicines/api/src/product.rs
```diff
@@ -1,9 +1,10 @@
...
+    substance::Substance,
+};
 use search_client::{models::IndexResult, Search};
 
-#[derive(GraphQLObject, Debug, Eq, Ord, PartialEq, PartialOrd)]
-#[graphql(description = "A medical product containing active ingredients")]
+#[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub struct Product {
     name: String,
     document_count: i32,
```
craiga, line 10: Why do we have document count here? Should be removed in favour of `totalCount` inside documents.

rjkerrison: the struct is no longer used to populate the graphql object – it's one of the nice advantages of using `#[graphql_object]` that we can decouple the type in rust from the fields sent to the user

PR #732, opened 2020-04-29
Add search_with_pagination_and_filter to search client

craiga:
Adds a `search_with_pagination_and_filter` to search client, required to support #621 and #401.


medicines/search-client/src/lib.rs
```diff
@@ -24,6 +24,17 @@ pub struct AzurePagination {
...
 }
 
+pub struct AzureFilterSet {
+    pub boolean_operator: String,
+    pub field_filters: Vec<AzureFieldFilter>,
+}
+
+pub struct AzureFieldFilter {
+    pub field_name: String,
+    pub operator: String,
```
sgrowe, line 34: It might be nice if `boolean_operator` and `operator` were `enum`s of the different things that are allowed and then we formatted them as strings as needed. Might make the interface a bit safer and nicer to use

PR #736, opened 2020-04-30
Document type filtering in API

craiga:
Adds the ability to filter by document type in the API.

Screenshot 2020-04-30 at 10 31 34

### DIFF
--- a/medicines/api/src/document.rs
+++ b/medicines/api/src/document.rs
@@ -1,6 +1,6 @@
 use crate::{pagination, pagination::PageInfo};
 use juniper::GraphQLObject;
-use search_client::{models::IndexResult, Search};
+use search_client::{models::IndexResult, AzureFieldFilter, AzureFilterSet, Search};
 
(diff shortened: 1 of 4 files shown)


medicines/api/src/document.rs
```diff
@@ -118,17 +121,47 @@ pub async fn get_documents(
...
+                    operator: "eq".to_string(),
+                    field_value: format!(
+                        "{}{}",
+                        document_type[..1].to_uppercase(),
+                        document_type[1..].to_lowercase()
+                    ),
+                })
+                .collect(),
+        },
+        None => AzureFilterSet {
```
rjkerrison, line 141: is it possible just to return `None` here instead? also – does this even work, or does azure freak out at being told to use `$filter=or`?

craiga: I've tested without any filters and it's working as expected.

I _could_ return `None` here, but that's going to require either:

 * `search_with_pagination_and_filter` on the search client to accept an `Option<AzureFilterSet>` which is a bit messy; or
 * an extra level of branching within `get_documents`.

I'm happy to implement either of those alternatives, but I feel like this is the neatest.

PR #743, opened 2020-04-30
Adding document type filter to products.

craiga:
Adds document type filter to products.

Has the downside of making a ton of queries to Azure search service.


medicines/api/src/product.rs
```diff
@@ -1,28 +1,18 @@
...
+    document::{self, get_documents},
     substance::Substance,
 };
+use juniper::FieldResult;
 use search_client::{models::IndexResult, Search};
 
 #[derive(Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub struct Product {
     name: String,
-    document_count: i32,
```
craiga, line 10: Should keep this around so we don't have to make round trips to Azure for every product just to get the document count.

sgrowe:
Currently some queries can be quite inefficient as they can result in one Azure search API call to fetch the list of products, and then one for each product to fetch it's list of documents, but we can address that in a follow up PR

medicines/api/src/product.rs
```diff
@@ -56,12 +56,8 @@ pub fn handle_doc(document: &IndexResult, products: &mut Vec<Product>) {
                 .iter_mut()
                 .find(|product| document_product_name == &product.name);
 
-            match existing_product {
```
rjkerrison, line 59: seems unnecessary to remove this if we're planning to add back a version of it – we could use the filters in the `documents()` call to filter a pre-existing `documents` list on the `Product`

medicines/api/src/product.rs
```diff
@@ -32,19 +22,29 @@ impl Product {
...
-            Some(a) => a,
-            None => 0,
-        };
-
-        get_documents_graph_from_documents_vector(docs, offset, self.document_count)
+
+    async fn documents(
+        &self,
+        first: Option<i32>,
+        after: Option<String>,
```
rjkerrison, line 29: cannot merge this without `skip` unless the website is updated to send `after`

medicines/api/src/product.rs
sgrowe: Yeah, you'd still need to make some other changes though so I think we might as well just put it back later

PR #750, opened 2020-05-01
Reduce Azure calls in the API

rjkerrison:
# Fewer calls for a faster API

Let's use as few calls as possible to Azure. The goal is 0 per API request, but I'd settle for 1 for now.

Fixes extra calls since #743.

Making calls

### Acceptance Criteria

- [ ] fewer calls are made than without this PR
- [ ] calls to Azure get a tracing::debug message (otherwise how can we see how few we make)

### Testing information

Run the API, hit it, watch how many calls it makes to Azure.

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/api/src/schema.rs
+++ b/medicines/api/src/schema.rs
@@ -58,7 +58,7 @@ impl QueryRoot {
     ) -> FieldResult<Documents> {
         let offset = get_offset_or_default(skip, after, 0);
 
-        get_documents(
+        let docs = get_documents(
             &context.client,
             search.as_deref().unwrap_or(" "),
             first,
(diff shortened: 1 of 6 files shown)


medicines/api/src/schema.rs
```diff
@@ -88,3 +109,14 @@ impl SubscriptionRoot {}
...
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_base_64_decode() {
+        let encoded = base64::encode("1229".to_string());
+        assert_eq!(base_64_decode(encoded).unwrap(), 1230);
```
sgrowe, line 120: Do we really want decode to return a different number than what was encoded? Feel like the 1 should be added elsewhere?

rjkerrison: yeah maybe the function name should be a bit clearer that this isn't a direct decoding

PR #752, opened 2020-05-04
Timlee/learning deploy still broken

TimboTambo:
# Remove frozen lockfile argument

Site wasn't building properly - had previously thought this was due to unspecified path prefix in build step, but was actually the frozen lockfile flag that was causing issues, so removing across all learning site workflows.

### Acceptance Criteria

- [ ] _The first thing this PR must achieve_
- [ ] _The second thing this PR must achieve_
- [ ] _etc_

### Testing information

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/.github/workflows/learning-web-branch.yaml
+++ b/.github/workflows/learning-web-branch.yaml
@@ -28,7 +28,7 @@ jobs:
 
       - name: Install modules
         working-directory: learning/web
-        run: yarn install --frozen-lockfile
+        run: yarn run install --frozen-lockfile
 
       # - name: Run tests with coverage
       #   working-directory: learning/web
@@ -36,7 +36,7 @@ jobs:
 
       - name: Build
         working-directory: learning/web
-        run: yarn build
+        run: yarn run build
         env:
           PATH_PREFIX: "/"
 
--- a/.github/workflows/learning-web-branch.yaml
+++ b/.github/workflows/learning-web-branch.yaml
@@ -65,8 +65,8 @@ jobs:
       #     ROOT_URL_DOMAIN: localhost
 
       - name: Deploy learning web to static site in azure storage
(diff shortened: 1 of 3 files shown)


.github/workflows/learning-web-branch.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
         working-directory: learning/web
         run: yarn build
         env:
-          PATH_PREFIX: ""
+          PATH_PREFIX: "/"
```
rjkerrison, line 41: does this change make a difference?

.github/workflows/learning-web-branch.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
 
       - name: Install modules
         working-directory: learning/web
-        run: yarn install --frozen-lockfile
```
sgrowe, line 31: If [`--frozen-lockfile`](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-frozen-lockfile) is throwing an error it's cause the lockfile is out of date so just running `yarn` to update `yarn.lock` should fix it

.github/workflows/learning-web-branch.yaml
TimboTambo: Nope - it just makes it explicit where it might have been ambiguous before, and makes it consistent with the other workflow files.

TimboTambo:
For posterity, the actual issue was that `yarn build` was running `gatsby build`, to produce the "production" build. The following test steps in CI were then running `gatsby develop`, which was replacing the contents of the `public` dir with the dev build. By moving the build step to after the test steps the "production" build was preserved.

PR #755, opened 2020-05-04
Use an enum for document type instead of strings

sgrowe:
Currently document type is just stored as a string in a lot of places. This PR tidies things up to use the `DocumentType` enum throughout.


medicines/api/src/document_type.rs
```diff
@@ -0,0 +1,78 @@
...
+            "PAR" => Ok(Self::Par),
+            _ => Err(DocTypeParseError {
+                source: s.to_string(),
+            }),
+        }
+    }
+}
+
+impl Display for DocumentType {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
```
rjkerrison, line 37: This all seems familiar – I think we have similar in `doc-index-updater`. Given that it's all about reading from `IndexResult`, and we're doing this in every crate where we read `IndexResult`, could we move this `DocumentType` into `search-client`?

medicines/api/src/product.rs
```diff
@@ -89,38 +89,40 @@ impl Product {
...
+) -> Result<Substance, anyhow::Error> {
     let azure_result = client
         .filter_by_collection_field("substance_name", substance_name)
         .await?;
 
+    let docs = azure_result
+        .search_results
+        .into_iter()
+        .map(TryInto::try_into)
+        .collect::<Result<Vec<Document>, _>>()?;
```
rjkerrison, line 121: so if one of them fails to parse, we fail completely?

medicines/api/src/document.rs
```diff
@@ -10,34 +15,47 @@ pub struct Document {
...
+    type Error = DocTypeParseError;
+
+    fn try_from(r: IndexResult) -> Result<Self, Self::Error> {
+        Ok(Self {
             product_name: r.product_name,
             active_substances: Some(r.substance_name),
             title: Some(r.title),
             created: r.created,
-            doc_type: Some(r.doc_type),
+            doc_type: Some(r.doc_type.parse()?),
```
rjkerrison, line 50: Given that `doc_type` is `Option`, I believe [we can use `.ok()`][ok] here and we'd be able to implement `From` instead of `TryFrom`:

```suggestion
            doc_type: r.doc_type.ok(),
```

[`result.ok()`][ok] is like a reverse of [`option.ok_or(error)`][ok or], both of them seem really useful!

[ok]: https://doc.rust-lang.org/std/result/enum.Result.html#method.ok
[ok or]: https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or

medicines/api/src/substance.rs
```diff
@@ -28,78 +32,83 @@ impl Substance {
...
+        });
+
+    for (substance, product, doc) in documents {
+        match substances.get_mut(&substance) {
+            None => {
+                let mut map = BTreeMap::new();
+                map.insert(product, vec![doc]);
+                substances.insert(substance, map);
+            }
+            Some(map) => match map.get_mut(&product) {
```
rjkerrison, line 77: can `map` be called `products` or `products_map` or `products_for_substance` or something, `map` is a bit vague and at first it reads to me like a method name

medicines/api/src/substance.rs
```diff
@@ -28,78 +32,83 @@ impl Substance {
...
+
+            let product = doc.product_name()?.to_string();
+
+            Some((substance, product, doc))
+        });
+
+    for (substance, product, doc) in documents {
+        match substances.get_mut(&substance) {
+            None => {
+                let mut map = BTreeMap::new();
```
rjkerrison, line 73: can `map` be called `products` or `products_map` or `products_for_substance` or something, `map` is a bit vague and at first it reads to me like a method name

medicines/api/src/substance.rs
```diff
@@ -28,78 +32,83 @@ impl Substance {
...
+                .substances()
+                .find(|s| s.starts_with(letter))?
+                .to_string();
+
+            let product = doc.product_name()?.to_string();
+
+            Some((substance, product, doc))
+        });
+
+    for (substance, product, doc) in documents {
```
rjkerrison, line 70: can this be broken out into another method? this `format_search_results` is 52 lines long

medicines/search-client/src/document_type.rs
```diff
@@ -0,0 +1,89 @@
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Formatter};
+use std::str::FromStr;
+
+#[derive(Serialize, Deserialize, Debug, Copy, Clone, Eq, PartialEq, Ord, PartialOrd)]
+#[cfg_attr(feature = "graphql", derive(juniper::GraphQLEnum))]
```
rjkerrison, line 6: excellent 👏

PR #758, opened 2020-05-05
Terraform updates

sgrowe:
- Updates [azure provider to 2.8.0](https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/CHANGELOG.md#280-april-30-2020)
- Switches to using [`address_prefixes`](https://www.terraform.io/docs/providers/azurerm/r/subnet.html#argument-reference) instead of the now deprecated `address_prefix` argument for subnets
- Fixes bootstrapping issue where it seems the latest version of terraform is too clever for it's own good and knows what the value of `azurerm_kubernetes_cluster.cluster.default_node_pool[0].vnet_subnet_id` is before it's even created `azurerm_kubernetes_cluster.cluster` (because we tell it what the `vnet_subnet_id` should be). I've had to add a bit more indirection to get terraform to wait until the cluster is created before trying to read its `route_table_id`.

### DIFF
--- a/infrastructure/environments/dev/main.tf
+++ b/infrastructure/environments/dev/main.tf
@@ -1,5 +1,5 @@
 provider "azurerm" {
-  version = "=2.2.0"
+  version = "=2.8.0"
   features {}
 }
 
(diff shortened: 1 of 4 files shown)


sgrowe:
Yeah 👍  ran it this morning

Issue #375, opened 2020-02-14
PARs  - Upload a new PAR


pataruco:
## User want

As a Medical Writer in the licensing team
I would like to upload a new PAR
So that I can ensure that new PARs are always available on products.mhra.gov.uk

(Linked to #397 and #398.)

### Customer acceptance criteria

- [x] Medical writers can upload a PAR PDF file to the website 
- [x] Medical writers can enter the relevant metadata about the document, such as the product name, active substances and Licence number
- [x] The new document is available on the products.mhra.gov.uk site
- [x] The PAR can be linked to one or multiple products (which will have their own PL number)
- [x] Medical writers must be logged in to the site with their MHRA account to upload and will be redirected to the sign in page if they are not. The can trigger the normal microsoft azure AD sign in from the page

Following fields are captured:
- [x] Product name per product
- [x] Active substances per product
- [x] Licence number for each product
- [x] Strength for each product
- [x] Pharmaceutical dose form for each product

### Technical acceptance criteria

- [x] PAR pdf is in blob storage after upload
- [x] PAR metadata is attached to blobs
- [x] PAR is in the Azure search index after upload
- [x] blob/metadata/search index is handled by doc-index-updater
- [x] Users accessing the app must be authenticated by being in the MHRA Azure group

### Infrastructure acceptance criteria

- [x] **terraform**: new blob storage container for temporary PAR storage
- [x] **terraform**: app registration
- [x] **terraform**: private blob container for pars-upload website
- [x] **terraform**: CDN / IP restriction / URL rewrite with SAS token
- [x] **github action**: master
   - [x] pars-upload push to non-prod private blob container
   - [ ] purges the non-prod cache of the CDN (unnecessary if we have HTML cache off)
- [x] **github action**:  branch
   - [x] doc-index-updater/pars-upload compatibility tests
   - [x] usual a11y stuff and cypress tests for pars-upload site 

 ### Data acceptance criteria

 ### Testing acceptance criteria

**Size**
XL

**Value**

**Effort**

### Exit Criteria met 
- [x] Backlog
- [x] Discovery
- [x] DUXD
- [x] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate 

#813

rjkerrison:
@TimboTambo @KaylieGreen1607 does this need to be closed/marked won't do/moved on the Github board? I can't tell where it should go to align with the physical board.

rjkerrison:
[...]

PR #795, opened 2020-05-15
Temporary Blob Storage

rjkerrison:
# Temporary Blob Storage

Creates and uses a temporary blob storage container for #375.

a neverending blob

### Acceptance Criteria

- [x] Introduce a `StorageClient` trait to handle adding and removing files
- [x] Add a file to temporary storage when receiving a PARs request
- [x] Retrieve that file when picking up the message

### Testing information

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
+++ b/medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
@@ -1,6 +1,6 @@
 use super::{
     models::{StorageClientError, StorageFile},
-    storage_client_factory_with_a_slightly_different_error_type, GetBlob, StorageClient,
+    GetBlob, StorageClient,
 };
 use crate::create_manager::hash;
 use async_trait::async_trait;
@@ -9,12 +9,14 @@ use azure_sdk_core::{
     MetadataSupport,
 };
 use azure_sdk_storage_blob::Blob;
+use azure_sdk_storage_core::client::Client;
 use std::collections::HashMap;
 
 pub struct AzureBlobStorage {
-    container_name: String,
+    pub container_name: String,
     prefix: String,
     storage_account: String,
+    master_key: String,
 }
 
 impl Default for AzureBlobStorage {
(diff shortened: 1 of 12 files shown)


medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
```diff
@@ -0,0 +1,129 @@
...
+use std::collections::HashMap;
+
+pub struct AzureBlobStorage {
+    pub container_name: String,
+    prefix: String,
+    storage_account: String,
+    master_key: String,
+}
+
+impl Default for AzureBlobStorage {
```
rjkerrison, line 22: I think this is never used, and was originally implemented because clippy said it should be when we had a `new()` implementation. Try removing it unless we have a use case for it. If clippy complains we can turn that warning off.

medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
```diff
@@ -0,0 +1,129 @@
...
+    fn default() -> Self {
+        Self::temporary()
+    }
+}
+
+impl AzureBlobStorage {
+    pub fn temporary() -> Self {
+        let container_name =
+            std::env::var("STORAGE_CONTAINER").expect("Set env variable STORAGE_CONTAINER first!");
```
rjkerrison, line 31: decouple from `permanent()` below by switching the `STORAGE_CONTAINER` for `TEMPORARY_STORAGE_CONTAINER`, which we can keep to the same value for now, but we should change with a new container created by terraform asap

medicines/doc-index-updater/src/storage_client/mod.rs
```diff
@@ -1,54 +1,39 @@
...
+mod client;
+mod delete;
+mod get;
+pub mod models;
 
-impl BlobClient {
-    pub fn new(azure_client: Client) -> BlobClient {
-        BlobClient { azure_client }
-    }
+pub struct BlobResponse {
```
rjkerrison, line 12: move this into `models`

PR #808, opened 2020-05-18
Sftp client refactor

rjkerrison:
# Refactors the Sentinel SFTP Client

Sentinel's SFTP client isn't special; it's just a way of grabbing a file from temporary storage.

This refactor aligns it with the temporary blob storage from Azure, decoupling the create_manager from how the files are retrieved.

_n.b. could probably do a bit better, go a bit more **strategy pattern** with this, but not right now_

### Acceptance Criteria

- [ ] Refactor SFTP client to move it away from create_manager
- [ ] Rely on storage_client trait to get blobs from temporary clients

### DIFF
--- a/medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
+++ b/medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
@@ -55,17 +55,16 @@ impl AzureBlobStorage {
         match base64::decode(&self.master_key) {
             Ok(_) => Ok(
                 Client::new(&self.storage_account, &self.master_key).map_err(|e| {
-                    StorageClientError::ClientError {
-                        message: format!("Couldn't create storage client: {:?}", e),
-                    }
+                    StorageClientError::ClientError(format!(
+                        "Couldn't create storage client: {:?}",
+                        e
+                    ))
                 })?,
             ),
-            Err(e) => Err(StorageClientError::ClientError {
-                message: format!(
-                    "Couldn't decode master key to create storage client: {:?}",
-                    e
-                ),
-            }),
+            Err(e) => Err(StorageClientError::ClientError(format!(
+                "Couldn't decode master key to create storage client: {:?}",
+                e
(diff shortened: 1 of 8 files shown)


medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
```diff
@@ -55,17 +55,16 @@ impl AzureBlobStorage {
         match base64::decode(&self.master_key) {
             Ok(_) => Ok(
                 Client::new(&self.storage_account, &self.master_key).map_err(|e| {
-                    StorageClientError::ClientError {
-                        message: format!("Couldn't create storage client: {:?}", e),
-                    }
+                    StorageClientError::ClientError(format!(
+                        "Couldn't create storage client: {:?}",
+                        e
+                    ))
```
TimboTambo, line 61: Think this `Ok` condition code would read more clearly (reduced nesting) if it was extracted out to a short function that takes care of creating the client or returning the mapped error.

medicines/doc-index-updater/src/storage_client/sftp_client.rs
```diff
@@ -1,35 +1,30 @@
...
-pub enum SftpError {
-    #[error("A TCP error connecting to server. ({0:?})")]
-    TcpError(#[from] std::io::Error),
-    #[error("An SSH error connecting to server. ({0:?})")]
-    Ssh2Error(#[from] async_ssh2::Error),
-    #[error("File could not be retrieved on server")]
-    CouldNotRetrieveFile,
-    #[error(transparent)]
-    Other(#[from] anyhow::Error),
+struct SftpInfo {
```
TimboTambo, line 11: Maybe "SftpConfig" would be more descriptive here? I think that would make it's use within `SftpClient` clearer, IMO.

rjkerrison: agreed

PR #811, opened 2020-05-19
PARs upload UI - part 2

sgrowe:
### Done
- [x] Updates CORs config for PARs upload endpoint
- [x] Broke the `new_par.js` mega file down into smaller components
- [x] Adds a `Wizard` component for managing going between form pages (and repeating the same form page)
- [x] Adds a stub `ReviewSubmission` screen
- [x] Update the E2E tests so they run against a mock of the doc-index-updater (I toyed with idea of running the actual doc-index-updater in CI but having it access the dev storage container, service bus queue and redis seemed a bit dodgy)
- [x] Added [prettier](https://prettier.io/) for consistent code formatting
- [x] Update the fields the API accepts to match what is submitted

### Still to do
- Add buttons for deleting an already added product or substance from the form (see https://app.zeplin.io/project/5dd51ae21205c944f8c1d35b/screen/5ebbfcb92c78c074bef52070)
- Flesh out `ReviewSubmission` screen
- Add submission complete screen
- Add submission failed message/screen


medicines/doc-index-updater/src/pars_upload.rs
```diff
@@ -14,26 +14,35 @@ use std::collections::HashMap;
...
     filters::multipart::{FormData, Part},
+    http::{header, Method},
     Filter, Rejection, Reply,
 };
 
 pub fn handler(
     state_manager: StateManager,
 ) -> impl Filter<Extract = impl Reply, Error = Rejection> + Clone {
+    let cors = warp::cors()
+        .allow_any_origin() // TODO restrict to a specific domain once we know what it is
```
rjkerrison, line 25: can we keep the restriction to `localhost` until we have other environments to which to restrict it? could introduce the environment variable here

medicines/doc-index-updater/src/pars_upload.rs
```diff
@@ -63,15 +72,22 @@ async fn queue_pars_upload(
     job_id: Uuid,
     form_data: FormData,
     state_manager: impl JobStatusClient,
-) -> Result<(), SubmissionError> {
-    let (metadata, file_data) = read_pars_upload(form_data).await.map_err(|e| {
+) -> Result<(), Rejection> {
+    let (metadatas, file_data) = read_pars_upload(form_data).await.map_err(|e| {
         tracing::debug!("Error reading PARS upload: {:?}", e);
-        e
+        warp::reject::custom(e)
```
rjkerrison, line 78: could create a `From<SubmissionError> for Rejection` to drop these `.map_err`

medicines/doc-index-updater/src/pars_upload.rs
```diff
@@ -63,15 +72,22 @@ async fn queue_pars_upload(
...
 
-    let _ = check_in_document_handler(document, state_manager).await;
+    for metadata in metadatas {
+        let storage_file = add_file_to_temporary_blob_storage(job_id, &file_data)
+            .await
+            .map_err(warp::reject::custom)?;
+
+        let document = document_from_form_data(storage_file, metadata);
+
+        check_in_document_handler(document, &state_manager).await?;
```
rjkerrison, line 88: There's a problem here in that we've got multiple documents on the queue with the same job_id. It might be better to have an array of job ids, one for each product, so they can succeed or fail independently. Otherwise, we'd need to have an array of statuses per job id, or some other more complicated state management.

We're also not doing anything with the `JobStatusResponse` here and maybe we should.

medicines/doc-index-updater/src/pars_upload.rs
```diff
@@ -83,11 +99,10 @@ async fn upload_pars_handler(
...
 
-    let _ = tokio::join!(tokio::spawn(queue_pars_upload(
-        job_id,
-        form_data,
-        state_manager.clone(),
-    )),);
+    let span = tracing::info_span!("Queueing PARS upload", job_id = job_id.to_string().as_str());
+    let _enter = span.enter();
+
+    queue_pars_upload(job_id, form_data, state_manager).await?;
```
rjkerrison, line 105: This is a big change in design from before.

Queueing was happening asynchronously so we could give the user the `job_id` quickly, otherwise they'll have to wait for 1 blob storage write and _n_ service bus message pushes.

medicines/pars-upload/.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
+  "semi": false,
```
rjkerrison, line 2: nice

rjkerrison: would have been nice as a separate PR, though – lots of changes to comb through in this part-linting, part-functional review

medicines/pars-upload/src/review_submission.js
```diff
@@ -0,0 +1,20 @@
+import { Layout } from './layout'
+import { H1 } from './typography'
+import { Button } from './button'
+import { BackLink } from './back-link'
+
+export const ReviewSubmission = ({ steps, currentStepIndex, submit }) => {
+  const pageData = steps.slice(0, currentStepIndex)
+
+  console.log('XXX', pageData)
```
rjkerrison, line 9: `console.log`

medicines/pars-upload/src/wizard.js
```diff
@@ -1,5 +1,62 @@
...
+
+  const currentPage = steps[pageIndex]
+  const Component = currentPage.component
+
+  const setPageData = (pageData) =>
+    setSteps((steps) =>
+      steps.map((page, i) =>
+        i === pageIndex ? { ...page, data: pageData } : page
+      )
+    )
```
rjkerrison, line 17: I think this looks good but it's a complex state machine so might call you with a few questions :)

medicines/doc-index-updater/src/pars_upload.rs
```diff
@@ -69,49 +70,53 @@ fn document_from_form_data(storage_file: StorageFile, metadata: BlobMetadata) ->
...
+    let request_id = Uuid::new_v4();
 
-    let span = tracing::info_span!("Queueing PARS upload", job_id = job_id.to_string().as_str());
+    let span = tracing::info_span!("PARS upload", request_id = request_id.to_string().as_str());
     let _enter = span.enter();
 
-    queue_pars_upload(job_id, form_data, state_manager).await?;
+    tracing::debug!("Received PARS submission");
+
+    let job_ids = queue_pars_upload(form_data, state_manager).await?;
```
rjkerrison, line 112: This is better now that we're returning all the different job ids, however adding to the queue is still synchronous, so the user has to wait for _n_ temporary blob uploads before receiving feedback on submission.

medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
```diff
@@ -75,12 +74,13 @@ impl StorageClient for AzureBlobStorage {
     async fn add_file(
         &self,
         file_data: &[u8],
+        license_number: &str,
```
rjkerrison, line 77: This is quite a strong opinion for the `storage_client` to have: previously, it only new it was handling files with metadata, not that there'd be anything like a `license_number` relevant for that file.

medicines/doc-index-updater/src/pars_upload.rs
majikandy: can we use our cdn endpoint here (+localhost for dev)?   https://mhraparsdev.azureedge.net

medicines/pars-upload/.prettierrc
sgrowe: Yeah, I should have put this in a separate PR but I just snapped at one point and had to sort the formatting out 😄

medicines/doc-index-updater/src/storage_client/azure_blob_client.rs
sgrowe: Yeah, good point. We can probably adjust this interface to be a bit cleaner somehow

PR #812, opened 2020-05-19
App registration for pars upload

majikandy:
Terraform for the app registration for pars upload

### DIFF
--- a/infrastructure/modules/products/app-registrations.tf
+++ b/infrastructure/modules/products/app-registrations.tf
@@ -1,6 +1,6 @@
 resource "azuread_application" "pars-upload" {
   name                    = "pars-upload-${var.namespace}"
-  reply_urls              = ["http://${var.namespace}"] #don't think we need this
+  reply_urls              = ["http://localhost:3000", "https://${var.namespace}"] #this url needs to be correct for the environment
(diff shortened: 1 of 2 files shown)


infrastructure/modules/products/app-registrations.tf
```diff
@@ -0,0 +1,16 @@
+resource "azuread_application" "pars-upload" {
+  name                    = "pars-upload-${var.namespace}"
+  reply_urls              = ["http://localhost:3000", "https://${var.namespace}"] #this url needs to be correct for the environment
```
majikandy, line 3: @rjkerrison @anubha22  we should be able to know this url after we'd done the terraform deploy part for the website

rjkerrison: will we want different reply_urls for different environments?

majikandy: yes, we could put them all in one array but we might as well have that array come from an environmental variable instead so that dev can have localhost + the deployed dev url and non prod and prod can just have their own web url

majikandy: in fact I'm gonna do that now in this PR and we can replace the values when we know them

Issue #799, opened 2020-05-18
Restrict public access to Azure key vault


TimboTambo:
## User want

As a business
I want to ensure sensitive info is as secret as possible
So that I can ensure access is to authorised personnel only

## Acceptance Criteria

Production key vault is accessible to MHRA IP ranges only

## Data - Potential impact

**Size**
S

**Value**
M

**Effort**

### Exit Criteria met 
- [ ] Backlog
- [ ] Discovery
- [ ] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

PR #815, opened 2020-05-19
Keyvault in terraform

TimboTambo:
# Bring keyvault into terraform

Bring keyvault (and production keyvault-specific resource group) into terraform and add IP restrictions to production keyvault.

Addresses #799

### Acceptance Criteria

- [ ] Brings keyvaults and any other associated resources into Terraform control
- [ ] Restricts az access in production to MHRA IP's (not hard coded in config file, introduced via env variable)

### Testing information

- Initialize terraform state in non-prod with `terraform init`
- Run `terraform plan` - observe that no changes are proposed to the keyvault resource(s)
- If you want to test the IP restriction, change the `Allow` to `Deny` in the module definition within `non-prod/main.tf`, apply it and check access from the terminal is blocked.

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved


infrastructure/environments/dev/.env.example
```diff
@@ -2,6 +2,9 @@ TF_VAR_CLIENT_ID=
 TF_VAR_CLIENT_SECRET=
 
 TF_VAR_CLUSTER_ROUTE_DESTINATION_CIDR_BLOCKS='["10.0.0.0/16"]'
-TF_VAR_CLUSTER_ROUTE_NEXT_HOP=
+TF_VAR_CLUSTER_ROUTE_NEXT_HOP="10.0.0.0"
```
majikandy, line 5: 👍

infrastructure/environments/prod/variables.tf
```diff
@@ -33,4 +34,27 @@ variable "CLUSTER_ROUTE_NEXT_HOP" {
...
+}
+
+variable "KEYVAULT_NAME" {
+  description = "Name of keyvault where secrets stored"
+  default     = "mhra-prod"
+}
+
+variable "KEYVAULT_AUTHORISED_PERSON_IDS" {
+  type        = list(string)
+  description = "IDs of objects (people etc) to associate access policies for"
```
majikandy, line 53: as in the object id when you view their profile in azure ad?

infrastructure/modules/keyvault/main.tf
```diff
@@ -0,0 +1,71 @@
...
+  tenant_id           = data.azurerm_client_config.current.tenant_id
+  soft_delete_enabled = true
+
+  sku_name = "standard"
+
+  network_acls {
+    default_action = var.network_acls_default_action
+    bypass         = "AzureServices"
+    ip_rules       = var.access_CIDR
+  }
```
rjkerrison, line 16: so for prod, we're using this to restrict on IP? do we need that as well as the access_policy below?

infrastructure/modules/keyvault/main.tf
```diff
@@ -0,0 +1,71 @@
...
+
+  tags = {
+    environment = var.environment
+  }
+
+  dynamic "access_policy" {
```
rjkerrison, line 22: This access policy seems to be granting every permission to every authorised person.

Do we have an access policy for reading only, or maybe for writing only?

infrastructure/modules/keyvault/variables.tf
```diff
@@ -0,0 +1,23 @@
...
+}
+variable "name" {
+  description = "Name of the keyvault"
+}
+variable "access_CIDR" {
+  description = "CIDR range to whitelist for access to the keyvault"
+}
+
+variable "authorised_person_ids" {
+  description = "IDs of objects (people etc) to associate access policies for"
```
rjkerrison, line 18: maybe rather than calling people objects, this could say _persons (incl scripted users)_, if my understanding is correct here

infrastructure/environments/prod/variables.tf
TimboTambo: I'm not sure quite sure how you'd go about finding this ID for a given person in the Azure profile with our current level of access, but I imagine that's where you would find it. I'm not sure what the best way of sharing this list of ID's is without adding to source control, but you can find them by running `terraform plan` and seeing the object ID's that would be removed. Maybe it's fine for them to be public and I just add in to the variables file.

infrastructure/modules/keyvault/main.tf
TimboTambo: This is the current list of permissions assigned to each of us currently, I didn't see any reason to lock it down further for each of us.

infrastructure/modules/keyvault/main.tf
TimboTambo: Yeah - we need both. This rule is specific for non-portal (API) access, which we want to lock down to MHRA IP's in line with pen testing suggestion. The below access policies refer to what each of us has permissions to do, and represents our current configuration that we set-up through the portal.

infrastructure/modules/keyvault/variables.tf
TimboTambo: All people are objects to me!
I updated wording with suggested text.

PR #818, opened 2020-05-20
Auto populate reply urls to the CDN address

majikandy:
- Automatically get the CDN url for the reply address for non prod and prod. Prod will also need the dns name when we get it.
 - Add localhost:3000 for dev but nowhere else

### DIFF
--- a/infrastructure/modules/cluster/diagnostics.tf
+++ b/infrastructure/modules/cluster/diagnostics.tf
@@ -16,8 +16,8 @@ resource "azurerm_storage_account" "logs" {
 resource "azurerm_monitor_diagnostic_setting" "cluster" {
   count              = var.log_cluster_diagnostics ? 1 : 0
   name               = var.diagnostic_setting_name
-  target_resource_id = "${azurerm_kubernetes_cluster.cluster.id}"
-  storage_account_id = "${azurerm_storage_account.logs[0].id}"
+  target_resource_id = azurerm_kubernetes_cluster.cluster.id
+  storage_account_id = azurerm_storage_account.logs[0].id
 
   dynamic "log" {
     for_each = var.diagnostic_log_types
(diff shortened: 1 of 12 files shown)


infrastructure/modules/cluster/diagnostics.tf
```diff
@@ -16,8 +16,8 @@ resource "azurerm_storage_account" "logs" {
 resource "azurerm_monitor_diagnostic_setting" "cluster" {
   count              = var.log_cluster_diagnostics ? 1 : 0
   name               = var.diagnostic_setting_name
-  target_resource_id = "${azurerm_kubernetes_cluster.cluster.id}"
-  storage_account_id = "${azurerm_storage_account.logs[0].id}"
```
majikandy, line 20: slipped this warning fix in too

infrastructure/modules/products/app-registrations.tf
```diff
@@ -1,6 +1,6 @@
 resource "azuread_application" "pars-upload" {
   name                    = "pars-upload-${var.namespace}"
-  reply_urls              = var.pars_reply_urls
+  reply_urls              = concat(["https://${azurerm_cdn_endpoint.pars.host_name}"], var.add_local_pars_reply_url ? ["http://localhost:3000"] : [])
```
majikandy, line 3: This is the main part for review - has already been applied to dev and works.

infrastructure/modules/keyvault/main.tf
```diff
@@ -2,7 +2,7 @@ data "azurerm_client_config" "current" {}
 
 resource "azurerm_key_vault" "secrets_vault" {
   name                = var.name
-  location            = var.location
+  location            = "uksouth"
```
majikandy, line 5: changed to hardcoded so that it didn't destroy and recreate when applied

infrastructure/modules/products/pars.tf
```diff
@@ -18,7 +18,7 @@ resource "azurerm_storage_account" "pars" {
 
 resource "azurerm_cdn_profile" "pars" {
   name                = "mhrapars${var.environment}"
-  location            = var.location
+  location            = "westeurope"
```
majikandy, line 21: not allowed CDN in uk south so needed to force to westeurope for non prod / prod (dev currently has default region as westeurope)

PR #825, opened 2020-05-21
Workflows for Pull Requests

rjkerrison:
# Workflows run for PRs

We have an issue with the `push` PRs only evaluating the last push, so in effect many broken builds could be included in a PR.

We'd like to run workflows for PRs instead, so that if any commit introduced by a PR has touched a certain file, the relevant workflows are run, and any failing build blocks the PR form merging.

a flag for PR

### Acceptance Criteria

- [ ] Workflows run based on the changeset in a PR
- [ ] As an example, this PR has checks for all the workflows changed, not just those in the most recent push

### Testing information

Check the workflow checks below.

### DIFF
--- a/.github/workflows/doc-index-updater-branch.yaml
+++ b/.github/workflows/doc-index-updater-branch.yaml
@@ -1,10 +1,9 @@
 name: doc-index-updater-branch
 
 on:
-  push:
+  pull_request:
     branches:
-      - "**"
-      - "!master"
+      - "master"
     paths:
       - medicines/doc-index-updater/**
       - medicines/search-client/**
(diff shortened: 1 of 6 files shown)


.github/workflows/doc-index-updater-branch.yaml
```diff
@@ -1,10 +1,9 @@
 name: doc-index-updater-branch
 
 on:
-  push:
+  pull_request:
     branches:
-      - "**"
-      - "!master"
+      - "master"
```
TimboTambo, line 6: Might we not be interested in running tests on pull requests to any branch?

rjkerrison: I did think about that, but in general not so much. If branches are short-lived and getting incremental changes into master is what we're striving for, then in my view PRs into branches are suggestions to a parent master-bound PR, and so those suggestions are not worth policing and it might even hinder development.

PR #828, opened 2020-05-21
temporary storage container

rjkerrison:
# Temporary storage container

Uses the temporary storage container for flow. 

Has equivalent changes in MHRA/deployments#30.

### Acceptance Criteria

- [x] use a different `.env` variable for temporary storage container
- [x] match the name created in terraform


rjkerrison:
Yeah, I don't like the naming much but I wanted to keep it close to the other one alphabetically. There might be something better we can do since terraform has these values hardcoded anyway.

PR #829, opened 2020-05-21
Show auth popup flow error in the government style

majikandy:
ezgif com-optimize

### DIFF
--- a/medicines/pars-upload/src/pages/_app.js
+++ b/medicines/pars-upload/src/pages/_app.js
@@ -22,10 +22,12 @@ function App({ Component, pageProps }) {
 
       if (error.includes('Login_In_Progress')) {
         error =
-          'Login popup is already open in another window, perhaps it is behind this window.'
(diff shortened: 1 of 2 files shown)


medicines/pars-upload/src/pages/_app.js
```diff
@@ -8,9 +8,25 @@ import { signIn, getAccount } from '../auth/authPopup'
...
+    if (error.startsWith('ClientAuthError:')) {
+      error = error.slice(16)
+
+      if (error.includes('Login_In_Progress')) {
+        error =
+          'Login popup is already open in another window, perhaps it is behind this window.'
+      }
+
+      setAuthError(error)
+    }
```
sgrowe, line 29: I think if there's an unknown error we want to show an error message still. Perhaps at the bottom we could add something roughly along the lines of 

```suggestion
    } else {
      setAuthError("Unknown error")
    }
```

majikandy: fair

PR #832, opened 2020-05-21
Fix cypress for pars upload

rjkerrison:
# Fix cypress for pars upload

- [x] running the mock as part of the test-e2e
- [x] cypress tests run in ci for pars upload

hey arnold

### DIFF
--- a/medicines/pars-upload/Makefile
+++ b/medicines/pars-upload/Makefile
@@ -19,12 +19,4 @@ set-env: ## Takes your current .env file and replaces the keyvault value with th
 
 .PHONY: e2e
 e2e: ## Runs end-to-end tests, including dev and mock servers, and cleans up afterwards.
-	yarn mock & wait-on http://localhost:8000; \
-	mock_server_pid = $$!; \
-	NEXT_PUBLIC_DISABLE_AUTH=true yarn dev & wait-on http://localhost:3000; \
-	dev_server_pid=$$!; \
-	test_exit_code=$$?; \
-	yarn run cypress run; \
-	kill $${dev_server_pid}; \
-	kill $${mock_server_pid}; \
-	exit $${test_exit_code}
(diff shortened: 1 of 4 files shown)


sgrowe:
Ah, the branch worflow [already runs `yarn mock`](https://github.com/MHRA/products/blob/master/.github/workflows/pars-upload-branch.yaml#L68) (outside of the `test-e2e` command). So we should probably either update the master workflow to run the same commands or remove the separate `yarn mock` command from the branch workflow.

rjkerrison:
Thanks @sgrowe and @majikandy. I've tried a different approach where I take those lengthy commands into the `Makefile` so that we have `make e2e` in both of the workflows. What do you think?

medicines/pars-upload/Makefile
```diff
@@ -16,3 +16,15 @@ set-env: ## Takes your current .env file and replaces the keyvault value with th
...
 	--name pars-env \
 	--file .env
+
+.PHONY: e2e
+e2e: ## Runs end-to-end tests, including dev and mock servers, and cleans up afterwards.
+	yarn mock & wait-on http://localhost:8000; \
+	mock_server_pid = $$!; \
+	NEXT_PUBLIC_DISABLE_AUTH=true yarn dev & wait-on http://localhost:3000; \
+	dev_server_pid=$$!; \
+	test_exit_code=$$?; \
```
sgrowe, line 26: I think this line should go after `yarn run cypress run` to pick up the exit code for the cypress test run, I think currently it's picking up the exit code of `wait-on http://localhost:3000` above?

rjkerrison: ah yes! I've reverted back to `yarn test-e2e` being the source of truth, with this `make e2e` being a pointer to it, so that should be fixed now

PR #834, opened 2020-05-21
PARs form success and error screens

### LINKED ISSUE #822: PARs - Upload success & error messages
Had a chat with @roughprada and it was decided:
- on successful form submission, there shoud be a success screen as per [the design](https://app.zeplin.io/project/5dd51ae21205c944f8c1d35b/screen/5ebbfcbb5f5cec74857f3c70)
- if the there's an error submitting the form, an [error summary component](https://design-system.service.gov.uk/components/error-summary/) should be shown above the page heading on the "Check your answers" page  (this matches the GOV.UK design system)
- no loading screen (for the time being at least), just submit the form when they click "Accept and send" on the "Check your answers" page and then either show them the success screen or add an error summary to the top of the page and scroll the user up to the top so they can see it.
- Update the HTML title in the browser tab when an error is thrown (as per GDS guidelines).

sgrowe:
Closes #822


.github/workflows/pars-upload-branch.yaml
```diff
@@ -10,7 +10,9 @@ on:
 
 env:
   NEXT_PUBLIC_DISABLE_AUTH: "true"
-  PARS_UPLOAD_URL: "http://localhost:8000"
+  PARS_UPLOAD_URL: "http://localhost:8000/pars"
+  # above is for cypress, below is for next.js
```
rjkerrison, line 14: if we're gonna separate, we could use the `CYPRESS_*` pattern to save passing them in, but doesn't have to block this PR

.github/workflows/pars-upload-master.yaml
```diff
@@ -10,7 +10,9 @@ on:
 
 env:
   NEXT_PUBLIC_DISABLE_AUTH: "true"
-  PARS_UPLOAD_URL: https://doc-index-updater.test.mhra.gov.uk/pars
+  PARS_UPLOAD_URL: "https://doc-index-updater.test.mhra.gov.uk/pars"
+  # above is for cypress, below is for next.js
+  NEXT_PUBLIC_PARS_UPLOAD_URL: "https://doc-index-updater.test.mhra.gov.uk/pars"
```
rjkerrison, line 15: do we have to pass this into the `.env` we create for builds?

PR #843, opened 2020-05-26
Temporarily fix auth not working in dev env

sgrowe:
This is a temporary hack to get submissions working in the dev env while we wait for MHRA to enable implicit granting of ID tokens

### DIFF
--- a/medicines/pars-upload/src/auth/authPopup.js
+++ b/medicines/pars-upload/src/auth/authPopup.js
@@ -1,16 +1,42 @@
 import { UserAgentApplication } from 'msal'
 import { msalConfig, loginRequest } from './authConfig'
 
-export function getAccount() {
-  const myMSALObj = new UserAgentApplication(msalConfig)
-  const account = myMSALObj.getAccount()
+export async function getAccount() {
+  const msalInstance = new UserAgentApplication(msalConfig)
+  const account = msalInstance.getAccount()
 
   if (account) {
+    // TODO:
+    //
+    // Temporary hack whilst we wait for implicit granting of ID tokens to be
+    // enabled by MHRA admins so we get an access token in the proper way:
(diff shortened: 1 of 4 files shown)


medicines/pars-upload/src/auth/authPopup.js
```diff
@@ -1,20 +1,46 @@
...
-  const myMSALObj = new UserAgentApplication(msalConfig)
-  const account = myMSALObj.getAccount()
+export async function getAccount() {
+  const msalInstance = new UserAgentApplication(msalConfig)
+  const account = msalInstance.getAccount()
 
   if (account) {
+    // TODO:
+    //
+    // Temporary hack whilst we wait for implicit granting of ID tokens to be
```
majikandy, line 11: To me this doesn't make sense, we have the token, why is it a hack to use it?

We do almost the same as in the documentation below apart from not using the `silent` approach

medicines/pars-upload/src/auth/authPopup.js
```diff
@@ -1,20 +1,46 @@
...
+    // enabled by MHRA admins so we get an access token in the proper way:
+    // https://github.com/AzureAD/microsoft-authentication-library-for-js/tree/dev/lib/msal-core#3-get-an-access-token-to-call-an-api
+
+    const token = window.sessionStorage['msal.idtoken']
```
majikandy, line 15: this only works if the setting in authConfig is set to session - if that is changed to local or cookies, this will break.

medicines/pars-upload/src/auth/authPopup.js
```diff
@@ -1,20 +1,46 @@
...
+      token,
       signOut: () => {
-        myMSALObj.logout()
+        msalInstance.logout()
       },
     }
   }
 }
 
+async function getToken(msalInstance) {
```
majikandy, line 28: is this unused code ?

PR #844, opened 2020-05-26
dev pars upload url

rjkerrison:
PARs Upload URL in dev environment will point to dev version of doc-index-updater. This aligns with the `infrastructure/scripts/update-hosts.sh` script's understanding of the URLs for the dev environment.

### DIFF
--- a/.github/workflows/doc-index-updater-branch.yaml
+++ b/.github/workflows/doc-index-updater-branch.yaml
@@ -10,7 +10,7 @@ on:
       - .github/workflows/doc-index-updater-branch.yaml
 
 env:
-  IMAGE: mhraproductsnonprodregistry.azurecr.io/products/doc-index-updater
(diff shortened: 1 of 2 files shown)


.github/workflows/doc-index-updater-branch.yaml
```diff
@@ -10,7 +10,7 @@ on:
       - .github/workflows/doc-index-updater-branch.yaml
 
 env:
-  IMAGE: mhraproductsnonprodregistry.azurecr.io/products/doc-index-updater
+  IMAGE: mhraproductsdevelopmentregistry.azurecr.io/products/doc-index-updater
```
majikandy, line 13: did you intent to use the dev registry? I thought we were only going to use the non-prod one ?

rjkerrison: I did intend this. I don't remember that, although it sounds reasonable as the tags shouldn't clash. If we're intending to only use non-prod one, should we remove the existing development one?

PR #850, opened 2020-05-27
access token

rjkerrison:
# Use the access token

Previously we were sending a `Bearer [object Object]` as the token. That's not right.


medicines/pars-upload/src/auth/authPopup.js
```diff
@@ -6,7 +6,8 @@ export async function getAccount() {
   const account = msalInstance.getAccount()
 
   if (account) {
-    const token = await getToken(msalInstance)
+    const token = (await getToken(msalInstance)).accessToken
+    console.log({ token })
```
sgrowe, line 10: TODO: remove

PR #853, opened 2020-05-27
Lint PARs upload project using ESLint

sgrowe:
Adds ESLint to check for basic code and accessibility errors


medicines/pars-upload/src/auth/authPopup.js
```diff
@@ -22,22 +22,6 @@ export async function getAccount() {
...
-    return await msalInstance.acquireTokenSilent(tokenRequest)
-  } catch (e) {
-    if (e.name === 'InteractionRequiredAuthError') {
-      return await msalInstance.acquireTokenPopup(tokenRequest)
-    } else {
-      throw e
-    }
-  }
-}
-
```
TimboTambo, line 40: Did you intend to include this?

rjkerrison: ah that function has been disused, so I guess the linter itself called out the cleanup

sgrowe: Yeah, exactly. 

Also the other changes were due to ESLint errors as well. All the button-related stuff is because the accessibility plugin was saying not to use `<a>` tags with `onClick` handlers unless you do a load of other stuff to make them accessible (which buttons automatically have it turns out)

PR #862, opened 2020-05-28
Terraform make apply and replyurls for non prod

majikandy:
- [x] _extend Make to do terraform apply_
- [x] _set reply urls to have primary web host as well as cdn_
- [x] _temporarily allow non prod to work with whichever of the 3 app registrations we configure._

### DIFF
--- a/infrastructure/environments/prod/main.tf
+++ b/infrastructure/environments/prod/main.tf
@@ -39,13 +39,14 @@ resource "azurerm_resource_group" "keyvault" {
 module "products" {
   source = "../../modules/products"
 
-  environment         = var.ENVIRONMENT
-  location            = var.REGION
-  namespace           = local.namespace
-  pars_namespace      = local.pars_namespace
-  resource_group_name = data.azurerm_resource_group.products.name
-  search_sku          = "standard"
-  app_registration_owners = var.KEYVAULT_AUTHORISED_PERSON_IDS
+  environment                       = var.ENVIRONMENT
+  location                          = var.REGION
+  namespace                         = local.namespace
+  pars_namespace                    = local.pars_namespace
+  resource_group_name               = data.azurerm_resource_group.products.name
+  search_sku                        = "standard"
+  app_registration_owners           = var.KEYVAULT_AUTHORISED_PERSON_IDS
+  addtional_allowed_pars_reply_urls = ["https://pars.mhra.gov.uk"]
 }
 
 data "azurerm_route_table" "load_balancer" {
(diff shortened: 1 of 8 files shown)


infrastructure/modules/products/app-registrations.tf
```diff
@@ -1,8 +1,7 @@
 resource "azuread_application" "pars-upload" {
   name                    = "pars-upload-${var.environment}"
-  reply_urls              = concat(["https://${azurerm_cdn_endpoint.pars.host_name}"], var.add_local_pars_reply_url ? ["http://localhost:3000"] : [])
+  reply_urls              = concat(["https://${azurerm_cdn_endpoint.pars.host_name}", "https://${azurerm_storage_account.pars.primary_web_host}", "https://mhraparsnonprod.z33.web.core.windows.net", "https://mhraparsnon-prod.azureedge.net"], var.add_local_pars_reply_url ? ["http://localhost:3000"] : [])
   group_membership_claims = "SecurityGroup"
```
TimboTambo, line 4: It would be better to pass in hardcoded, environment-specific values as var's.

majikandy: I'm starting to think we just hardcode (or environment var) the lot and have one app registration

rjkerrison: @majikandy if we do that, could we include it in each environment's `main.tf`, so no matter which environment we're terraforming, it checks that the app registration is there?

majikandy: Agreed with @TimboTambo we will keep the separate app registrations just incase we want to try something with the app registration and no affect prod app registration.

So probably the happy medium is that we have one for all non prod and one for prod.

infrastructure/environments/prod/main.tf
```diff
@@ -39,13 +39,14 @@ resource "azurerm_resource_group" "keyvault" {
...
-  search_sku          = "standard"
-  app_registration_owners = var.KEYVAULT_AUTHORISED_PERSON_IDS
+  environment                       = var.ENVIRONMENT
+  location                          = var.REGION
+  namespace                         = local.namespace
+  pars_namespace                    = local.pars_namespace
+  resource_group_name               = data.azurerm_resource_group.products.name
+  search_sku                        = "standard"
+  app_registration_owners           = var.KEYVAULT_AUTHORISED_PERSON_IDS
+  addtional_allowed_pars_reply_urls = ["https://pars.mhra.gov.uk"]
```
majikandy, line 49: assuming this is the dns name we'll get ?

rjkerrison: should we allow CDN and static container ones too?

majikandy: they are, but they are in the module directly

rjkerrison: I see them, `azurerm_cdn_endpoint.pars.host_name` and `azurerm_storage_account.pars.primary_web_host`

PR #863, opened 2020-05-28
Fix second submission of the form errorring

sgrowe:
When you click “Accept and Send” on the review page the click event handler mistakenly sets the click event as the “data” for that page, so when you hit send again it tries to include the click event in the form data and throws a `TypeError`

### DIFF
--- a/medicines/pars-upload/src/review_submission.js
+++ b/medicines/pars-upload/src/review_submission.js
@@ -63,7 +63,12 @@ export const ReviewSubmission = ({
         }
       })}
 
-      <Button type="button" onClick={submit}>
+      <Button
+        type="button"
+        onClick={() => {
+          submit(null)
+        }}
+      >
         Accept and send
       </Button>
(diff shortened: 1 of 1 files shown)


medicines/pars-upload/src/review_submission.js
```diff
@@ -63,7 +63,12 @@ export const ReviewSubmission = ({
         }
       })}
 
-      <Button type="button" onClick={submit}>
+      <Button
+        type="button"
+        onClick={() => {
+          submit(null)
```
rjkerrison, line 69: This has potential gotcha written all over it if someone removes the `null` later. Maybe we could have a test that submits twice?

sgrowe: Yeah, that's definitely a good idea. If I have time once the other PARs stuff is done I'll try and get round to it but if not I'll just merge this as is 😅

PR #864, opened 2020-05-28
Dynamic reply url

majikandy:
.


TimboTambo:
A test or two would be quite nice...

majikandy:
> A test or two would be quite nice...

Just around the url split? Makes sense

medicines/pars-upload/src/auth/authPopup.js
```diff
@@ -2,11 +2,11 @@ import { UserAgentApplication } from 'msal'
 import { msalConfig, loginRequest } from './authConfig'
 
 export async function getAccount() {
+  msalConfig.auth.redirectUri = getCurrentHost(window.location.href)
```
sgrowe, line 5: Would it make sense to put this bit in `authConfig.js` instead?

medicines/pars-upload/src/auth/authPopup.js
```diff
@@ -22,7 +22,13 @@ export async function getAccount() {
   }
 }
 
+export function getCurrentHost(currentFullUrl) {
+  var uriComponents = currentFullUrl.split('/')
+  return uriComponents[0] + '//' + uriComponents[2]
+}
+
 export function signIn() {
+  msalConfig.auth.redirectUri = getCurrentHost(window.location.href)
```
sgrowe, line 31: ^ Same as above 

> Would it make sense to put this bit in `authConfig.js` instead?

medicines/pars-upload/src/auth/authPopup.js
majikandy: I had it there initially  but it couldn't access `window`

PR #868, opened 2020-05-29
Fix typo: "License" -> "Licence"

sgrowe:
In British English licen**s**e is the verb and licen**c**e is the noun.

Source: https://www.grammarly.com/blog/licence-license/

### DIFF
--- a/medicines/doc-index-updater/src/pars_upload.rs
+++ b/medicines/doc-index-updater/src/pars_upload.rs
@@ -46,11 +46,11 @@ pub fn handler(
 async fn add_file_to_temporary_blob_storage(
     _job_id: Uuid,
     file_data: &[u8],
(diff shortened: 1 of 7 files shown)


medicines/doc-index-updater/src/pars_upload.rs
```diff
@@ -221,7 +221,7 @@ fn product_form_data_to_blob_metadata(
     let product_names = vec![product_name];
 
     let title = get_field_as_string(&fields, "title")?;
-    let pl_number = get_field_as_string(&fields, "license_number")?;
```
sgrowe, line 224: The interface between the frontend and the API has slightly changed by renaming this field but I've run the E2E tests locally (without the mock server) to make sure it still works fine

PR #870, opened 2020-05-29
Capitalise PARs upload field values

### LINKED ISSUE #860: Field values should be converted to ALL CAPS by API
Allow capital letters and non-capital letters in fields but all results should be capitalised when processed

See point 2 in https://github.com/MHRA/products/issues/845

sgrowe:
Allows capital letters and non-capital letters as input in each field but then uppercases them before saving.

I've also put the code specifically for handling `FormData` into it's own module to tidy things up a bit.

Closes #860


medicines/doc-index-updater/src/pars_upload.rs
```diff
@@ -2,21 +2,17 @@ use crate::{
     create_manager::models::BlobMetadata,
     document_manager::{accept_job, check_in_document_handler},
     models::{Document, FileSource},
+    multipart_form_data::*,
```
TimboTambo, line 5: How come * here instead of specifically named structs/functions?

sgrowe: Just laziness. Have updated 😄

Issue #801, opened 2020-05-18
Restrict outbound traffic from VNET


TimboTambo:
## User want

As a business 
I want to ensure that egress from the cluster is to only the necessary services
So that if compromised, the possible implications are limited

## Acceptance Criteria

Only whitelisted external service endpoints/IPs can be reached from inside a cluster pod.

**Size**
M

**Value**
S

**Effort**

### Exit Criteria met 
- [ ] Backlog
- [ ] Discovery
- [ ] DUXD
- [ ] Development 
- [ ] Quality Assurance 
- [ ] Release and Validate

PR #948, opened 2020-08-03
Add network security group for cluster

TimboTambo:
# Add network security group for cluster

Relates to #801 

Restrict inbound and outbound traffic from cluster subnets using network security groups.

### Acceptance Criteria

- [ ] Restrict inbound and outbound connections to the cluster, whilst ensuring existing required connections are allowed

### Testing information

Test hitting doc-index-updater and medicines-api endpoints and ensuring documents can still be processed and uploaded.

### Pre-flight Checklist

- [ ] DUXD review approval
- [ ] Accessibility Reviewed
- [ ] Product owner demo

### Pre-merge Checklist

- [ ] Branch test approved

### DIFF
--- a/infrastructure/environments/dev/main.tf
+++ b/infrastructure/environments/dev/main.tf
@@ -1,5 +1,5 @@
 provider "azurerm" {
-  version = "=2.8.0"
+  version = "=2.10.0"
   features {}
 }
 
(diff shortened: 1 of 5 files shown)


TimboTambo:
Resolutions will need to be fixed before this can be merged.
Also I think it would be good to add egress rules from the cluster subnet based not on destination (which requires hardcoded IP addresses, which we don't have for e.g. github), but based on source IP, so that the request should originate from the cluster IPs (e.g. .67 and .68), since we know the cluster uses it's own egress restrictions that are based on url.

PR #1050, opened 2020-10-20
Manifests

gilbertsak:
The purpose of this PR is to make this repo the source of manifests for kubernetes and istio.


StuartHarris (approved):
we did this as a group, so all good

PR #1054, opened 2020-10-28
Object path vulnerability

TimboTambo:
# Object path vulnerability

- Update nextjs to version 10 to remove vulnerability in object-path dependency, flagged by dependabot.
- Update node to v14 due to incompatibility of some libraries with v13
- Add env to release workflows over discovered issue that snapshots won't match in production
- Don't use dev server for cypress testing in pars-upload app in order to avoid possible hot reloading issue which reloads page and causes flakiness


StuartHarris (approved):
reviewed together

PR #1066, opened 2020-12-21
Timlee/new pl formats

TimboTambo:
# Add support for new PL prefixes

Add support for new PL-type prefixes

### Acceptance Criteria

- [x] When a message is sent to the Doc Index Updater with any of the PL prefixes (PL, PLNI, PLGB, THR, NR), it is saved in normalized form in Azure (e.g. PLNI123451234)
- [x] When a user searches for a document using any of the PL prefixes (PL, PLNI, PLGB, THR, NR) and the site is using Azure, the correct documents are returned 
- [x] When a user searches for a document using any of the PL prefixes (PL, PLNI, PLGB, THR, NR) and the site is using the GraphQL API, the correct documents are returned


gilbertsak (approved):
Yeah it looks good, Tim has also gone through the changes through teams call. Pending the automatic checks go through, the changes look ok.

PR #1071, opened 2021-01-12
Gsak/waitonaxios

gilbertsak:
This is a pull request to fix security axios 0.19.2 to axios@^0.21.1. 

axios is a plugin only used by wait-on. Wait-on is used to wait for files, ports, sockets, http(s) resources. Where exactly it is used is not 100% known. The changes where minor versions and accord to yarn 🧶. No update here should be a breaking one.

I did: yarn upgrade wait-on@5.2.1   (was from 5.2.0)
in web and pars-upload directories 

this change  updated decencies:
axios@0.21.1
├─ follow-redirects@1.13.1
├─ joi@17.3.0

### DIFF
--- a/medicines/pars-upload/package.json
+++ b/medicines/pars-upload/package.json
@@ -49,6 +49,6 @@
     "pa11y-ci": "^2.4.0",
     "prettier": "^2.1.2",
     "react-test-renderer": "^16.13.1",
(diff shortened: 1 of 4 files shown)


spriceware (approved):
Reviewed changes.

PR #1078, opened 2021-03-11
Release workflow no image

TimboTambo:
# Release workflow no image

The current release workflows for `doc-index-updater` and `medicines-api` work from a tagged commit. Currently the workflow will fail if there was no image generated for that commit - this happens when only the manifests have been  updated. This makes it currently impossible to release changes only to the manifests in production.

This PR adds some conditions to continue with the workflow when the image does not exist (i.e. when the `Fetch image for tagged commit` step fails).

### DIFF
--- a/.github/workflows/doc-index-updater-branch.yaml
+++ b/.github/workflows/doc-index-updater-branch.yaml
@@ -79,14 +79,18 @@ jobs:
           cargo build --release
           cargo test --release
 
+      - name: Create Docker tag
+        working-directory: products/medicines/doc-index-updater
+        run: |
+          TAG="$(git rev-parse --short=7 ${{ github.sha }})"
+          echo "TAG=$TAG" >>$GITHUB_ENV
+
       - name: Build and push Docker image
         if: steps.filter.outputs.src == 'true'
         working-directory: products/medicines/doc-index-updater
         run: |
-          TAG="$(git rev-parse --short=7 ${{ github.sha }})"
           make ci-master tag=$TAG image=$IMAGE
           DIGEST="$(docker inspect --format='{{index .RepoDigests 0}}' ${IMAGE}:${TAG})"
-          echo "TAG=$TAG" >>$GITHUB_ENV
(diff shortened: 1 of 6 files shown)


StuartHarris (approved):
This looks great to me. Thanks @TimboTambo

.github/workflows/doc-index-updater-branch.yaml
```diff
@@ -79,11 +79,16 @@ jobs:
...
+          echo "TAG=$TAG" >>$GITHUB_ENV
+
       - name: Build and push Docker image
         if: steps.filter.outputs.src == 'true'
         working-directory: products/medicines/doc-index-updater
         run: |
-          TAG="$(git rev-parse --short=7 ${{ github.sha }})"
           make ci-master tag=$TAG image=$IMAGE
           DIGEST="$(docker inspect --format='{{index .RepoDigests 0}}' ${IMAGE}:${TAG})"
           echo "TAG=$TAG" >>$GITHUB_ENV
```
TimboTambo, line 94: Can remove this line in each of them as line 86 tackles this!

.github/workflows/doc-index-updater-release.yaml
```diff
@@ -28,6 +28,8 @@ jobs:
           password: ${{ secrets.REGISTRY_PASSWORD }}
 
       - name: Fetch image for tagged commit
+        id: fetch-image
+        continue-on-error: true
         working-directory: ./products/medicines/doc-index-updater
         run: |
           TAG="$(git rev-parse --short=7 $GITHUB_SHA)"
```
TimboTambo, line 35: The release workflows need the Docker step added too

.github/workflows/doc-index-updater-branch.yaml
pataruco: Good catch, I missed

.github/workflows/doc-index-updater-release.yaml
pataruco: done

StuartHarris (approved):
Awesome! Much better as a separate step.

PR #1079, opened 2021-03-23
Terraform update

TimboTambo:
# Terraform update

Update to Terraform 14.

### DIFF
--- a/infrastructure/modules/service-bus/output.tf
+++ b/infrastructure/modules/service-bus/output.tf
@@ -1,7 +1,13 @@
 output "queues_default_primary_connection_string" {
-  value = ["azurerm_servicebus_queue_authorization_rule.service_bus_queue_auth_rule.*.primary_connection_string"]
+  value = tomap({
+    for k, v in azurerm_servicebus_queue_authorization_rule.service_bus_queue_auth_rule :
+    k => v.primary_connection_string
+  })
 }
 
 output "queues_default_primary_key" {
-  value = ["azurerm_servicebus_queue_authorization_rule.service_bus_queue_auth_rule.*.primary_key"]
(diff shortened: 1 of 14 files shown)


infrastructure/modules/service-bus/output.tf
```diff
@@ -1,7 +1,7 @@
 output "queues_default_primary_connection_string" {
-  value = ["${azurerm_servicebus_queue_authorization_rule.service_bus_queue_auth_rule.*.primary_connection_string}"]
+  value = ["azurerm_servicebus_queue_authorization_rule.service_bus_queue_auth_rule.*.primary_connection_string"]
```
StuartHarris: I think I made a mistake here. The quotes should probably be removed. As it stands the output is just the string (i.e. not resolved):
```
service_bus_queue_keys = [
  "azurerm_servicebus_queue_authorization_rule.service_bus_queue_auth_rule.*.primary_key",
]
```

infrastructure/modules/service-bus/output.tf
```diff
@@ -1,7 +1,13 @@
 output "queues_default_primary_connection_string" {
-  value = ["${azurerm_servicebus_queue_authorization_rule.service_bus_queue_auth_rule.*.primary_connection_string}"]
+  value = tomap({
```
StuartHarris, line 2: also, because we are generating the resources using `for_each` now, the resources are not a list, but rather a map and so we need to iterate over them map to produce outputs

StuartHarris (commented):
our service bus outputs weren't quite right so I updated them, and tested with `terraform refresh`

StuartHarris (approved):
I think this looks good now.

PR #1088, opened 2021-04-09
Turn off Graph QL temp

gilbertsak:
Checking search results with GraphQL on non-prod only. Will revert back later if nothing change


spriceware (approved):
Hi. Yes, think it would be a good idea to investigate on non-prod.

PR #1092, opened 2021-04-14
Fetch existing digest

TimboTambo:
# Fetch existing digest if not being updated

If no Docker image is generated/found for the commit in the workflows (because there are no changes to the app code), take the image digest that is already in the deployment manifests.

This fixes a bug where the digest in the Deployments repository would otherwise be overwritten with the old digest from the manifests in this (Products) repository.

### DIFF
--- a/.github/workflows/doc-index-updater-release.yaml
+++ b/.github/workflows/doc-index-updater-release.yaml
@@ -82,16 +82,22 @@ jobs:
         with:
           kustomize: 3.4.0
           command: |
-            set -eux
+            set -ex
             SOURCE="${PWD}/products/manifests/doc-index-updater/overlays/prod"
             DEST="${PWD}/deployments/doc-index-updater/prod"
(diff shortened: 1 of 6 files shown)


.github/workflows/doc-index-updater-release.yaml
```diff
@@ -82,16 +82,22 @@ jobs:
         with:
           kustomize: 3.4.0
           command: |
-            set -eux
+            set -ex
```
StuartHarris, line 85: in all these examples, I think we should keep `set -u` in here if we can to protect against unset variables (just changing it to `+u` and back to `-u` around the smallest bits possible)

StuartHarris (changes requested):

.github/workflows/doc-index-updater-release.yaml
TimboTambo: Ah agreed! I'd mixed up +u and -u for this line.

StuartHarris (approved):

PR #1093, opened 2021-04-14
turned GraphQL back on

gilbertsak:
Turning GraphQL back on, the issue was the docker image was the incorrect digest. So no need for this to be off


TimboTambo (approved):

gilbertsak:
azure_search_tests.js failed it's final step in uploading document in checks. Not sure if that is how it is meant to be but that is unaffected by this change.

PR #1098, opened 2021-04-21
updated web release yaml to have full commit SHA number

gilbertsak:
Fix the issue where the GIT action wont complete because the commit SHA is not the full number.


spriceware (approved):
Looks good

spriceware (commented):
Thanks. Added full SHA

PR #1099, opened 2021-04-29
add full SHA to pars-upload yaml files

gilbertsak:
Update the pars-upload yaml files to have full SHA as the GIT action fails without this.


spriceware (approved):
Thanks Gilbert. looks good

PR #1105, opened 2021-06-17
Added PLGB and PLNI options to drop down pars upload dropdown

gilbertsak:
Added PLGB and PLNI options to drop down. 
Two files needed to be updated it looks like:
products_form.js
update_products_form.js 

Will test using https://mhraparsnonprod.azureedge.net/ and give to UAT team to test. 

	Acceptance criteria:
1.	I can upload a PAR with a licence number that has a PLNI prefix
2.	I can upload a PAR with a licence number that has a PLGB prefix
3.	I can update a PAR to have a licence number that has a PLNI prefix
4.	I can update a PAR to have a licence number that has a PLGB prefix
5.	The external user sees the correct licence number prefix assigned to the PAR


spriceware (approved):
Nice. See you have added the two additional options. Thanks

PR #1124, opened 2021-11-02
Add frontend UI for territory filtering

### LINKED ISSUE #1123: Add radio buttons to UI
Ensure that focus is returned to the filters on the left hand side after submitted.

siame:
We want to be able to filter products by the territories they're 
approved in.

The method we're using to filter territories is similar to how we're
currently filtering documents.

We've added the filter checkboxes to the left hand side of the page,
and when each filter is selected it is appended to a query string in the
page's URL.

The implementation of this territory filtering has been mostly
generalising the current implementation of filtering documents. We're
then able to re-use most of the existing functionality by passing in the
existing `DocType` or new `TerritoryType`.

Closes #1123.

For the snapshot fixes I ran `yarn jest -u` to allow jest to update the
snapshots for us, Otherwise the querystring-interpreter tests just
needed to modify a failing import.

### DIFF
--- a/medicines/web/src/components/search-filter/index.tsx
+++ b/medicines/web/src/components/search-filter/index.tsx
@@ -1,7 +1,11 @@
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useRef, useState } from 'react';
 import styled from 'styled-components';
 import { RerouteType } from '../../model/rerouteType';
-import { DocType } from '../../services/azure-search';
+import {
+  DocType,
+  SearchType,
+  TerritoryType,
+} from '../../services/azure-search';
 import { Button, Checkbox } from '../form-elements';
 
 const StyledSearchFilter = styled.section`
@@ -19,13 +23,14 @@ const StyledSearchFilter = styled.section`
     }
 
     label {
-      flex: 1;
+      display: flex;
+      align-items: center;
     }
   }
 `;
 
 const ButtonContainer = styled.div`
-  padding-top: 10px;
+  padding-top: 15px;
 `;
 
 const Fieldset = styled.fieldset`
@@ -36,6 +41,12 @@ const Fieldset = styled.fieldset`
 const Legend = styled.legend`
   display: block;
   font-size: 18px;
+  font-weight: bold;
+  margin-top: 19px;
+`;
+
+const SubLegend = styled.legend`
+  display: block;
   font-weight: bold;
   margin: 19px 0;
 `;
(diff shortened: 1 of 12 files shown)


medicines/web/src/components/search-filter/index.tsx
```diff
@@ -49,78 +60,103 @@ const AccessibleHeading = styled.h3`
...
+      enabledFilters.splice(filterIndex, 1);
     } else {
-      enabledDocTypes.push(docTypeToToggle);
+      enabledFilters.push(filterToToggle);
     }
-    setCheckedFilters(enabledDocTypes);
+    setCheckedFilters(enabledFilters);
   };
 
   const submit = (e) => {
```
gilbertsak, line 157: When pressing the submit button the checked filters does not maintain for "Applicable to territory" when the page is refreshed (the tick disappears of what you have selected).  Doesn't look like this was intended at the moment looking at the code?

gilbertsak (commented):
When pressing the submit button the checked filters does not maintain for "Applicable to territory" when the page is refreshed (the tick disappears of what you have selected). Doesn't look like this was intended at the moment looking at the code?

medicines/web/src/components/search-filter/index.tsx
siame: I'm having some trouble replicating this, I'm currently seeing the below:

https://user-images.githubusercontent.com/9799940/139842195-8f0c55c9-a414-4b7d-814e-25e76a3039e0.mov 

Could you detail the steps you used to get the filters not persisting?

gilbertsak: I see the issue I think. So search for a product say "paracetamol" then do what you did. When you choose a product from the list using A -  Z etc. it works for me but not when searching for a product.

siame: 👍 got it, I see what the issue is now, will put out a fix shortly.

siame: @gilbertsak Should be fixed in the latest commit

gilbertsak:
Ok that works nice! 

Another similar issue (that might not be an issue, if I am expecting something I shouldn't be) - In URL I don't see the "&ter=UK" etc added to URL on search page (after searching paracetamol) when I am selecting just only the new filters and clicking submit. I know it not meant to do anything at the moment but it doesn't appear in URL because I do see the value in the URL when using A - Z items and conducting same test.

siame:
@gilbertsak Yep, I can see what was casuing that issue in the last commit - I've just amended that in this latest one so that should be fixed here too.

gilbertsak (approved):
looks all good to me

medicines/web/src/components/search-filter/index.tsx
```diff
@@ -138,9 +174,17 @@ const SearchFilter: React.FC<ISearchFilterProps> = (props) => {
...
+        {generateDocTypeCheckboxFor(
+          DocType.Spc,
+          'Summary of Product Characteristics',
+        )}
+        {generateDocTypeCheckboxFor(DocType.Pil, 'Patient Information Leaflet')}
+        {generateDocTypeCheckboxFor(DocType.Par, 'Public Assessment Reports')}
+        <SubLegend ref={filterHeader}>Applicable to territory</SubLegend>
+        {generateTerritoryTypeCheckboxFor(TerritoryType.UK, 'United Kingdom')}
+        {generateTerritoryTypeCheckboxFor(TerritoryType.NI, 'Northern Ireland')}
+        {generateTerritoryTypeCheckboxFor(TerritoryType.GB, 'Great Britain')}
```
TimboTambo, line 187: Did you/MHRA decide against using radio buttons?

TimboTambo (approved):

PR #1125, opened 2021-11-03
Add territory to doc-index-updater and search

siame:
Now that `territory` has been added as a field in Sentinel, we want to
process this value when passed into the `doc-index-updater`.

This commit updates both the `doc-index-updater` and `search` projects.

For `doc-index-updater`, we've updated the `Document` struct with
`territory`, and all of its dependent code references and tests. The
`audit_logger`, `create_manager` and `delete_manager` should now all be
aware of this new field.

For `search` we've added `territory` to the default search index. We've
also added some better error handling for the methods that communicate
with Azure's REST interface, most notably using PUT requests rather than
POST requests to create and update indexes.


StuartHarris (approved):
we paired, and reviewed together

medicines/search/src/main.rs
```diff
@@ -18,7 +18,7 @@ async fn main() {
             .expect("Failed to delete datasource"),
         ("create_index", Some(m)) => {
```
TimboTambo, line 19: I think it would make sense to update the command to be `create_or_update_index` too, updating the `cli.yaml` and the `README.md` accordingly, but get that we're working to a tight schedule so will approve these changes and leave it to you to decide.

TimboTambo (dismissed):

medicines/search/src/main.rs
siame: Good shout, will update that.

StuartHarris (approved):
nice

PR #1126, opened 2021-11-03
Link territory changes from UI to GraphQL backend

siame:
This PR links together the prior changes from #1124 and #1125 so that we can now filter territories in the UI, have the backend parse our new GraphQL query, then query and filter in our Azure search service on our newly uploaded territory field.

Below is a short summary of what each commit does in this PR.

---

### Read territory as TerritoryType not String

I originally modified `doc-index-updater` with the `territory` field as
a String, much like `pl_number`.

After working through the project more, I've found that the `territory`
field is more suited to being an Enum, like `doc_type`. This gives us
both the safety that we're only parsing certain values within this
field, as well as allowing de/serialisation restrictions when indexing.

### Read and parse territoryType in graphql query 

This commit adds the functionality to filter by territoryType when
receiving this filter over graphql.

We're currently only filtering by territory, not by pl_number prefix.

### Connect territory filtering from UI to backend

This commit sends the territory filters to the backend when using the
graphql query mode. This allows us to filter by the territory field in
the metadata.

### Change territory to be read as TerritoryType 
This commit also adds better error handling when receiving an index
update. We now error on incorrect TerritoryTypes the same way we error
on incorrect DocumentTypes.

### DIFF
--- a/medicines/search-client/src/document_type.rs
+++ b/medicines/search-client/src/document_type.rs
@@ -55,59 +55,6 @@ impl Display for DocumentType {
     }
 }
 
-#[cfg_attr(
-    feature = "graphql",
-    async_graphql::Enum(desc = "Territory type (UK/GB/NI)"),
-    derive(Serialize, Deserialize, Debug, Ord, PartialOrd)
-)]
-#[cfg_attr(
-    not(feature = "graphql"),
-    derive(
-        Serialize,
-        Deserialize,
-        Debug,
-        Copy,
-        Clone,
-        Eq,
-        PartialEq,
-        Ord,
-        PartialOrd
-    )
-)]
-pub enum TerritoryType {
-    #[serde(alias = "Uk")]
-    UK,
-    #[serde(alias = "Gb")]
-    GB,
-    #[serde(alias = "Ni")]
-    NI,
-}
-
-impl FromStr for TerritoryType {
-    type Err = TerritoryTypeParseError;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        match s.to_ascii_uppercase().as_str() {
-            "UK" => Ok(Self::UK),
-            "GB" => Ok(Self::GB),
-            "NI" => Ok(Self::NI),
-            _ => Err(TerritoryTypeParseError {
-                source: s.to_string(),
-            }),
-        }
-    }
-}
-
-impl Display for TerritoryType {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        match self {
-            TerritoryType::UK => write!(f, "UK"),
-            TerritoryType::GB => write!(f, "GB"),
-            TerritoryType::NI => write!(f, "NI"),
-        }
-    }
-}
-
 #[derive(Debug, Clone)]
 pub struct DocTypeParseError {
     source: String,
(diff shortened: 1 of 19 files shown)


gilbertsak (approved):
I can see nothing wrong with this myself, the failed check on git was talking about not seeing the cookie banner. So maybe git has cached the choice?  This wouldn't of been part of this change anyway.

medicines/search-client/src/document_type.rs
```diff
@@ -55,6 +55,59 @@ impl Display for DocumentType {
     }
 }
 
+#[cfg_attr(
```
TimboTambo, line 58: May make more sense to have this in a new file (`territory_type/rs`?) or rename this file (can't think of a name right now that would cover both doc type and territory type).

medicines/search-client/src/document_type.rs
```diff
@@ -96,11 +166,38 @@ mod tests {
...
+
+        assert_eq!(territory_type, as_str);
+    }
+
+    #[test_case("\"UK\"", TerritoryType::UK; "uk uppercase")]
+    #[test_case("\"Uk\"", TerritoryType::UK; "uk titlecase")]
+    #[test_case("\"GB\"", TerritoryType::GB; "gb uppercase")]
+    #[test_case("\"Gb\"", TerritoryType::GB; "gb titlecase")]
+    #[test_case("\"NI\"", TerritoryType::NI; "ni uppercase")]
+    #[test_case("\"Ni\"", TerritoryType::NI; "ni titlecase")]
```
TimboTambo, line 195: Guessing there are no lower-case requirements?

medicines/search-client/src/models.rs
```diff
@@ -171,14 +173,14 @@ impl From<IndexResult> for IndexEntry {
...
-                Some(t) => t,
-                None => "".to_owned(),
-            },
             title: res.title,
             pl_number: vec![],
             file_name: res.file_name,
             doc_type: res.doc_type,
+            territory: match res.territory {
+                Some(t) => t,
+                None => TerritoryType::UK,
```
TimboTambo, line 182: I can't remember how the `IndexEntry` is used - what's the implication of defaulting to UK here instead of making it optional on `IndexEntry` and defaulting to None?

TimboTambo (dismissed):

medicines/search-client/src/document_type.rs
siame: Will move this across. Lucky for me it's exposed through `models.rs` to the other projects in this repo so don't even have to change all of the imports 😅 .

medicines/search-client/src/document_type.rs
siame: AFAIK it's uppercase only, but i've added titlecase to be on par with how the `doc_type` is parsed. It's not outside the realms of possibility for it to be lowercase too, so I'll add that in.

medicines/search-client/src/models.rs
siame: Good shout here, will send this through as an `Option`.

siame: Found a few other places where this was used as a fixed value rather than parsed from incoming XML and have addressed this in the latest commit, so real good spot with this 👍 .

TimboTambo (approved):

PR #1127, opened 2021-11-04
Derive territory on import

siame:
Azure's search doesn't allow us to prefix search the `pl_number` field.
Therefore in order to search for a territory through the `pl_number`
prefix, we derive the `territory` from the `pl_number` during import,
but only if a `territory` is not already given.

The basic rules are:

    If territory is present in document:
      return territory
    Else:
      If pl_number prefix for known territory type:
        return matching territory type
      Else:
        return None


TimboTambo (approved):
Important that MHRA are aware of the implications in terms of sending through those files without territory again in order for the filtering behaviour to work as expected, and our recommendation for them to do a one-time update on their side of all the files to correct the data.

StuartHarris (approved):
worked on this together

TimboTambo (approved):

PR #1146, opened 2022-02-01
Change order of prefixes into occurrence-order

siame:
PL prefixes are the most uploaded documents, followed by THR, followed
by NR. This commit changes the order of the items in the PAR
upload/update dropdown to reflect this.


gilbertsak (approved):
All good

PR #1238, opened 2022-09-05
update front notice text email

gilbertsak:
Front page message update


steve-price-mhra (approved):
Hi Gilbert. Approving now

PR #1240, opened 2022-09-06
updated text

gilbertsak:
Updated text


steve-price-mhra (approved):
Approving. Thanks Gilbert

PR #1245, opened 2022-11-18
Testing graphql turn of flag works

gilbertsak:
Testing graphql flag


gilbertsak:
Then we can see if no more calls to graphql at https://mhraproductsnonprod.z33.web.core.windows.net/

PR #1251, opened 2022-12-13
Reverting back to true to use graphQL to see if master branch works with release int…

gilbertsak:
GraphQL back to true to see if goes into static folder fine.


steve-price-mhra (approved):
Hi Gilbert, thanks. Just approving revert - Steve

PR #1254, opened 2022-12-13
Yaml file back to how it was with false GraphQL flag

gilbertsak:
Azure-Storage-Action not allowing action to run and deploy so will have to close change and return to this later.


steve-price-mhra (approved):
Hi Gilbert, thanks. Looks Ok, all back to disable GRAPHQL use

PR #1319, opened 2024-04-16
BMGF on flag

gilbertsak:
Change BMGF flag on workflow


steve-price-mhra (approved):
Reviewed proposed change to yaml. Thanks. Looks Ok

PR #1323, opened 2025-03-20
update font

gilbertsak:
Update font


steve-price-mhra (approved):
Yes, looks good to go with font changes

PR #1324, opened 2025-03-25
updating frontpage text

gilbertsak:
Updating front page text


steve-price-mhra (approved):
Thanks, can see the comment out. Ok to go

